PR #5, opened 2024-12-04
feat: Add EnsureHPSF function

TylerHelmuth:

- Add solution for easily checking if an input string is valid HPSF

- Add new `EnsureHPSF` to allow easily checking if a string is valid HPSF
- Update Validator interface to return `error` instead of `[]error` to make Validate work easier with Go's typical `if err == nil` pattern.  
- Keep the ability to report multiple errors at once using `errors.Join()`

### DIFF
--- a/pkg/hpsf/hpsfTypes.go
+++ b/pkg/hpsf/hpsfTypes.go
@@ -206,7 +206,7 @@ type HPSF struct {
 
 func (h *HPSF) Validate() error {
 	if h.Components == nil && h.Connections == nil && h.Containers == nil && h.Layout == nil {
-		return errors.New("default HPFS structs are considered invalid")
+		return errors.New("default HPSF structs are considered invalid")
 	}
 
 	results := []error{}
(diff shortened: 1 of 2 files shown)


pkg/hpsf/hpsfTypes.go
```diff
@@ -200,20 +204,35 @@ type HPSF struct {
...
 	}
 	for _, c := range h.Containers {
 		e := c.Validate()
-		results = append(results, e...)
+		results = append(results, e)
+	}
+	return errors.Join(results...)
+}
+
+// EnsureHPFS returns an error if the input is not HPFS yaml or invalid HPFS
```
codeboten, line 229:
```suggestion
// EnsureHPSF returns an error if the input is not HPSF yaml or invalid HPSF
```

pkg/hpsf/hpsfTypes.go
```diff
@@ -200,20 +204,35 @@ type HPSF struct {
...
+func EnsureHPFS(input string) error {
```
codeboten, line 230:
```suggestion
func EnsureHPSF(input string) error {
```

pkg/hpsf/hpsfTypes.go
```diff
@@ -200,20 +204,35 @@ type HPSF struct {
 	Layout      Layout       `yaml:"layout,omitempty"`
 }
 
-func (h *HPSF) Validate() []error {
+func (h *HPSF) Validate() error {
+	if h.Components == nil && h.Connections == nil && h.Containers == nil && h.Layout == nil {
+		return errors.New("default HPFS structs are considered invalid")
```
codeboten, line 209:
```suggestion
		return errors.New("default HPSF structs are considered invalid")
```

MikeGoldsmith (approved):
Looks good. I like the use of errors.join 👍🏻

PR #6, opened 2024-12-04
fix: Switch back to slice errors

TylerHelmuth:

- Switches the Validator interface back to []error 

- update the interface
- adjust interface implementations


kentquirk:
Doh, I was doing this too, but I'm also adding comments. Mind if we just close this in favor of mine (coming soon)?

PR #8, opened 2024-12-16
feat: add support for templated components

kentquirk:

- Adds support for components that are defined in a data file and use a Go template to generate config information.

- Create a couple of templated components
- Create the infrastructure that can load and render them
- Integrate into HPFS as a component type
- Extend the generation functions to accept a map of userdata
- Create a slimmer hpsf demo file (hpsf2)


pkg/config/components/honeycombExporter.yaml
```diff
@@ -0,0 +1,35 @@
+name: ToHoneycomb
+kind: HoneycombExporter
+summary: Export traces to Honeycomb in Honeycomb's event format
```
codeboten, line 3:
```suggestion
summary: Export telemetry to Honeycomb
```

pkg/config/components/honeycombExporter.yaml
```diff
@@ -0,0 +1,35 @@
...
+description: |
+  This component exports traces to Honeycomb in Honeycomb's event format.
```
codeboten, line 5:
```suggestion
  This component exports telemetry to Honeycomb.
```

pkg/config/components/honeycombExporter.yaml
codeboten: my suggestion here is that this component could be generic enough to emit any kind of data to honeycomb, the export format might not be necessary to call out in the documentation of the component.

pkg/config/components/honeycombExporter.yaml
```diff
@@ -0,0 +1,35 @@
...
+    direction: input
+    type: HoneycombTraces
+properties:
+  - name: API_Key
+    summary: The API key to use to authenticate with Honeycomb
+    description: |
+      The API key to use to authenticate with Honeycomb.
+    type: string
+    validations: [nonblank]
+    default: ${HONEYCOMB_TRACES_APIKEY}
```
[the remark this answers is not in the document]
kentquirk: This was deliberate to avoid confusion with API keys you may be using for other purposes; it's not a world where one honeycomb key makes sense anymore. I think it's likely that you'll have multiple Honeycomb keys, so I've changed the name to make it even clearer.

pkg/config/components/honeycombExporter.yaml
kentquirk: This component today only applies to Refinery, but it's a good point that logs can also flow through Refinery. So I changed it to telemetry but left in the "event format" part.

codeboten (approved):

PR #9, opened 2024-12-19
feat(hpsf): add ability to produce otlp exporter to honeycomb

codeboten:

This PR adds the ability to use the HoneycombExporter component template to produce collector configuration to emit otlphttp to honeycomb.

- adding template for collector_config
- added tests for both refinery and collector config

### DIFF
--- a/pkg/config/components/honeycombExporter.yaml
+++ b/pkg/config/components/honeycombExporter.yaml
@@ -23,6 +23,14 @@ properties:
     type: string
     validations: [nonblank, url]
     default: https://api.honeycomb.io
+  - name: Compression
+    summary: The compression to use when sending data to Honeycomb
+    description: |
+      The compression to use when sending data to Honeycomb.
+      This is normally gzip, but can be overridden.
+    type: string
+    validations: [nonblank]
+    default: gzip
 templates:
   - kind: refinery_config
     name: HoneycombExporter_RefineryConfig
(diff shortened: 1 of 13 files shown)


pkg/config/components/honeycombExporter.yaml
```diff
@@ -23,6 +23,14 @@ properties:
     type: string
     validations: [nonblank, url]
     default: https://api.honeycomb.io
+  - name: Compression
+    summary: The compression to use when sending data to Honeycomb
+    description: |
+      The compression to use when sending data to Honeycomb.
+      This is normally gzip, but can be overridden.
+    type: string
+    validations: [nonblank]
```
MikeGoldsmith, line 32: Does having this as `nonblank` mean you have to provide a value. Is `none` okay?

pkg/config/testdata/HoneycombExporter_output_collector_config.yaml
```diff
@@ -0,0 +1,10 @@
+API_Key: ${HONEYCOMB_TRACES_APIKEY}
+API_URL: https://api.honeycomb.io
```
MikeGoldsmith, line 2: Naming is always fun. There's 3 different forms here; `API_Key`, `AcessKeys` and `sendKey` :|

```suggestion
ApiKey: ${HONEYCOMB_TRACES_APIKEY}
ApiUrl: https://api.honeycomb.io
```

MikeGoldsmith (approved):
Looks good. Left a non-blocking query and suggestion.

kentquirk:
I think this is now obsolete and should be closed. Is that correct?

codeboten:
Yes this can now be closed, the functionality has been moved to different pr's

PR #10, opened 2025-01-06
feat: Add default configuration

MikeGoldsmith:

When creating pipelines based on HPSF, it's important to start with a simple and valid configuration. This PR adds a configuration that be used as a basis for initial pipeline configurations.

- Adds a simple default valid configuration that includes a simple Refinery configuration with a deterministic sampler

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -1,0 +1,16 @@
+package config
+
+// DefaultConfiguration is the default HPSF configuration that includes a
+// simple Refinery configuration with a determinisic sampler.
+const DefaultConfiguration = `
+components:
+  - name: DefaultDeterministicSampler
+    kind: DeterministicSampler
+    properties:
+      - name: Environment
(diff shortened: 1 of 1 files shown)


kentquirk:
Let's talk about this in sync. There are two issues:
* This isn't actually a valid configuration, because it doesn't have importers or exporters
* The current plan is to have multiple starting templates to choose from when you create a new configuration; we shouldn't need to define them in Go code because they'll be stored as YAML in a database.

I hear from this that we need a little more explanation of how this is all going to fit together. I'll share some docs.

kentquirk (approved):
The long term plan here is that users will choose from one of several templates to start, and so we won't be using a default configuration built into the go code at all -- we'll be fetching something from a database. But it seems like it might be useful to have this for now.

So as long as we understand that this is likely just going to go away, there's no harm in having it until we build that database.

PR #12, opened 2025-01-11
feat: a bunch of refactoring and beginnings of collector support

kentquirk:

- Ability to generate a collector config
- A couple of collector components
- Add a lot of comments
- Reorganize for easier maintenance

- Render is now RenderToMap
- The user's HPSF is now available to the template and used
- Some refinery-specific stuff is now more generic
- Additional config type for collector
- Remove the generated hash because it's not solving the right problem
- We should be using text/template not html/template
- Two collector components (that will be merged in a future PR)
- Lots of comments added

The code in this PR is functional and runs to completion, but it's not correct yet. It was just a plausible place to stop. #13 builds on this one.

### DIFF
--- a/pkg/config/component.go
+++ b/pkg/config/component.go
@@ -20,7 +20,7 @@ const (
 // the values are any valid YAML value.
 // We will need to convert the dotted paths into real ones later.
 type Component interface {
-	GenerateConfig(Type, map[string]any) (tmpl.TemplateConfig, error)
+	GenerateConfig(cfgType Type, userdata map[string]any) (tmpl.TemplateConfig, error)
 }
 
 type NullComponent struct{}
--- a/pkg/config/component.go
+++ b/pkg/config/component.go
@@ -1,6 +1,9 @@
 package config
 
-import "github.com/honeycombio/hpsf/pkg/config/tmpl"
+import (
+	"github.com/honeycombio/hpsf/pkg/config/tmpl"
+	"github.com/honeycombio/hpsf/pkg/hpsf"
+)
 
 type Type string
 
@@ -24,4 +27,29 @@ type NullComponent struct{}
 
 func (c NullComponent) GenerateConfig(Type, map[string]any) (tmpl.TemplateConfig, error) {
 	return nil, nil
+}
+
+// This base component is used to make sure that the config will be valid
+// even if it stands alone. This is likely to be a temporary solution until we have a
+// database of components.
+type GenericBaseComponent struct {
+	Component hpsf.Component
+}
+
+func (c GenericBaseComponent) GenerateConfig(ct Type, userdata map[string]any) (tmpl.TemplateConfig, error) {
+	switch ct {
+	case RefineryConfigType:
+		return tmpl.DottedConfig{
+			"General.ConfigurationVersion": 2,
+			"General.MinRefineryVersion":   "v2.0",
+		}, nil
+	case RefineryRulesType:
+		return tmpl.DottedConfig{
+			"RulesVersion": 2,
+		}, nil
+	case CollectorConfigType:
+		return tmpl.NewCollectorConfig(), nil
+	default:
+		return nil, nil
+	}
 }
--- a/pkg/config/component.go
+++ b/pkg/config/component.go
@@ -1,8 +1,6 @@
 package config
 
-import (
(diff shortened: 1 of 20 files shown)


pkg/config/component.go
```diff
@@ -19,11 +20,36 @@ const (
...
+			"General.MinRefineryVersion":   "v2.0",
+		}, nil
+	case RefineryRulesType:
+		return tmpl.DottedConfig{
+			"RulesVersion": 2,
+		}, nil
+	case CollectorConfigType:
+		return tmpl.NewCollectorConfig(), nil
+	default:
+		return nil, nil
```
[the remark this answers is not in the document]
kentquirk: No, although I should comment it. GenerateConfig returns nil if it can't generate a config of the requested type. It's specifically not an error because not all components are appropriate for all configs.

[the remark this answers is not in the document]
codeboten: right, but all components should fall into one of the other cases above no?

pkg/config/tmpl/collectorconfig.go
```diff
@@ -0,0 +1,154 @@
...
+
+// CollectorConfig is a config specifically focused on creating collector
+// YAML configuration files.
+type CollectorConfig struct {
+	Sections map[string]DottedConfig
+	// receivers  DottedConfig
+	// processors DottedConfig
+	// exporters  DottedConfig
+	// extensions DottedConfig
+	// service    DottedConfig
```
codeboten, line 17:
```suggestion
```

pkg/config/templateComponent.go
```diff
@@ -69,6 +71,25 @@ type TemplateComponent struct {
 	Properties  []TemplateProperty `yaml:"properties,omitempty"`
 	Templates   []TemplateData     `yaml:"templates,omitempty"`
 	User        map[string]any     `yaml:"user,omitempty"`
+	hpsf        *hpsf.Component    // the component from the hpsf document
+}
+
+// SetHPSF is making a copy here (by not taking a pointer argument) to make sure 
+// that the original doesn't get modified.
```
codeboten, line 78:
```suggestion
// SetHPSF stores the original component's details and may modify their contents. To
// prevent the original being modified, the argument here should never be changed to a pointer.
```

pkg/config/tmpl/dottedconfig.go
```diff
@@ -30,35 +29,54 @@ func (dc DottedConfig) renderInto(m map[string]any, key string, value any) {
...
-	h := md5.New()
-	hash := hex.EncodeToString(h.Sum(data))
-	return data, hash, nil
+	return data, nil
 }
 
-func (dc DottedConfig) Merge(other DottedConfig) DottedConfig {
-	for k, v := range other {
-		dc[k] = v
+// Merge merges two DottedConfigs together.
```
codeboten, line 51:
```suggestion
// Merge combines two `DottedConfig` structs together, the values from the
// `DottedConfig` passed in will override any values that are not slices.
```

pkg/config/components/OTelGRPCReceiver.yaml
```diff
@@ -0,0 +1,39 @@
+name: OTel_GRPC_Receiver
+kind: OTelGRPCReceiver
+summary: Receives OpenTelemetry gRPC traces
+description: |
+  Imports telemetry to Honeycomb from OpenTelemetry gRPC traces (not HTTP).
+ports:
```
codeboten, line 6: any thoughts on renaming this to `io` or `connections`.... i dont have any other alternatives... i thought of `connectors`, that might be acceptable if the collector concept is never bubbled up to this level

pkg/config/components/OTelHTTPReceiver.yaml
```diff
@@ -0,0 +1,39 @@
...
+properties:
+  - name: Host
+    summary: The hostname or IP address on which to listen
+    description: |
+      Hostname or IP address on which to listen for incoming traces.
+      It is recommended not to change the default unless
+      you know what you're doing.
+    type: string
+    validations: [nonblank]
+    default: ${STRAWS_COLLECTOR_POD_IP}
```
codeboten, line 19:
```suggestion
    default: ${COLLECTOR_POD_IP}
```

codeboten (approved):
just a few more minor suggestions

PR #13, opened 2025-01-11
feat: combine receivers, enable conditionals in templates

kentquirk:
This is based on #12 and should be merged after it.

- The previous collector support didn't know how to combine things that touched the same component. Now it does.

- Add dedup function for use in services section
- Allow components to have `suppress_if` along with `key` and `value` in the data section; it's also a template so the full capability of the template language is available for evaluating the decision.


pkg/config/templateComponent.go
```diff
@@ -62,13 +65,33 @@ type TemplateData struct {
...
+	connections []*hpsf.Connection `yaml:"connections,omitempty"`
+}
+
+// we're making a copy here to make sure it doesn't get modified
+func (t *TemplateComponent) SetHPSF(c hpsf.Component) {
+	t.hpsf = &c
+}
+
+// HProp is a template helper that gets a map of all properties specified in the hpsf document.
+func (t *TemplateComponent) HProps() map[string]any {
```
MikeGoldsmith, line 86: I'm not sure HProps is clear enough for what it is but also HPSFProps maybe feels cumbersome?

MikeGoldsmith (approved):
Looks good. I think the refactoring work really helps make things clearer and adds better structure around the different config types.

pkg/config/templateComponent.go
kentquirk: We definitely want these names to be short because they're in templates that can't easily be broken into multiple lines; I chose this because it's like Props but different. Let's live with it for a little while; it's not hard to change if we find it confusing.

PR #16, opened 2025-01-17
fix: remove duplicate collector section declarations

robbkidd:

- Correcting the emitted collector config.

- add a test for collector config output
  - currently depends on #15 for the addition of testify to go dependencies
- remove duplicate section from config key templating

Changes collector config output from ...

    receivers:
      receivers:
        otlp/x:

... to ...

    receivers:
      otlp/x:

- use the collector component Name property

Reference the instance property Name when rendering a component instance
instead of the name of the component type.

1. This is clearer in the rendered config. Items have the names
   humans gave them.
2. Avoids naming collisions for HPSF that have two or more of the
   same type of component.

Turns ...

    receivers:
      otlp/otlp_receiver:

... into ...

    receivers:
      otlp/otlp_in:

### DIFF
--- a/pkg/translator/translator_test.go
+++ b/pkg/translator/translator_test.go
@@ -1,0 +1,85 @@
+package translator
+
+import (
+	"strings"
+	"testing"
+
+	"github.com/honeycombio/hpsf/pkg/config"
+	"github.com/honeycombio/hpsf/pkg/hpsf"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	yamlv3 "gopkg.in/yaml.v3"
+)
+
+func TestGenerateConfig(t *testing.T) {
+	const inputData = `components:
+  - name: RefineryGRPC_2
+    kind: RefineryGRPC
+    ports:
+      - name: TraceOut
+        direction: output
+        type: Honeycomb
+    properties:
+      - name: Port
+        value: 4317
+        type: number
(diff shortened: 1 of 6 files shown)


codeboten (approved):

pkg/translator/translator_test.go
```diff
@@ -0,0 +1,85 @@
...
+  - source:
+      component: otlp_in
+      port: Traces
+      type: OTelTraces
+    destination:
+      component: otlp_out
+      port: Traces
+      type: OTelTraces`
+
+	const expectedConfig = `receivers:
```
codeboten, line 50: could the expected config live in `testdata/` file 😬

robbkidd: Absosmurfly. Are you picturing a testdata/ folder at the root of the repo?

kentquirk: > The simplest solution is to keep all the test data in the same directory as the code itself under a special folder called testdata. This is a special folder ignored by the compiler, but still accessible for reading during tests. 

**https://hackandsla.sh/posts/2020-11-23-golang-test-fixtures/#keeping-test-data-under-testdata**

robbkidd: Moved embedded HEREDOC(ish) strings to `./testdata/` files.

TylerHelmuth (approved):

PR #18, opened 2025-01-17
fix: remove incorrect otlp export config

TylerHelmuth:

- The collector config for otlpexporter we were generating was using the wrong schema. It doesn't need `protocols.grpc.`.

- Removed wrong schema
- updated test expectation


robbkidd (approved):
Yup

PR #19, opened 2025-01-17
fix: http export translation

robbkidd:

- Remove extraneous sub keys from the OTLP HTTP exporter, similar to #18 

- add a test case for the HTTP exporter
- make that test pass


robbkidd:
I don't know that we'll want to keep growing the test suite here for variations on component rendering, but it's a place to start.

TylerHelmuth (approved):

PR #20, opened 2025-01-22
feat: add component for collector debug export

robbkidd:

- Wanna debug that Collector, man.

- Added an `OTelDebugExporter` component definition and an integrationy test for it in `translator_test.go`.

### DIFF
--- a/pkg/config/components/OTelDebugExporter.yaml
+++ b/pkg/config/components/OTelDebugExporter.yaml
@@ -7,12 +7,9 @@ properties:
   - name: Verbosity
     summary: The verbosity level of the debug output
     description: |
-      The verbosity level of the debug output. The default is "basic".
+      The verbosity level of the debug output. Valid values are basic, normal, or detailed. The default is "basic".
     type: string
-    validations:
-      - basic
-      - normal
-      - detailed
+    validations: [nonblank]
     default: basic
 templates:
   - kind: collector_config
--- a/pkg/config/components/OTelDebugExporter.yaml
+++ b/pkg/config/components/OTelDebugExporter.yaml
@@ -1,0 +1,27 @@
+name: otel_debug_exporter
+kind: OTelDebugExporter
+summary: Exports pipeline signal traffic to stdout
+description: |
+  Exports signal data in a pipeline to stdout.
+properties:
+  - name: Verbosity
+    summary: The verbosity level of the debug output
+    description: |
+      The verbosity level of the debug output. The default is "basic".
+    type: string
+    validations:
+      - basic
+      - normal
+      - detailed
+    default: basic
+templates:
+  - kind: collector_config
+    name: otel_debug_exporter_collector
+    format: collector
+    meta:
+      componentSection: exporters
+      signalTypes: [traces, metrics, logs] # we'll generate a name for each pipeline if there's more than 1
+      collectorComponentName: debug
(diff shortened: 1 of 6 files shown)


pkg/config/components/OTelDebugExporter.yaml
```diff
@@ -0,0 +1,27 @@
+name: otel_debug_exporter
+kind: OTelDebugExporter
+summary: Exports pipeline signal traffic to stdout
+description: |
+  Exports signal data in a pipeline to stdout.
+properties:
+  - name: Verbosity
```
robbkidd, line 7: How much of the surface of the debug exporter config would we like to surface in this initial component version?

pkg/config/components/OTelDebugExporter.yaml
```diff
@@ -0,0 +1,27 @@
...
+    summary: The verbosity level of the debug output
+    description: |
+      The verbosity level of the debug output. The default is "basic".
+    type: string
+    validations:
+      - basic
```
robbkidd, line 13: I'm probably holding these validations wrong. The intent: declare a strict list of valid string values.

pkg/translator/testdata/simple_grpc_collector_config.yaml
```diff
@@ -6,11 +6,13 @@ receivers:
             http:
                 endpoint: ${STRAWS_COLLECTOR_POD_IP}:1234
 exporters:
+    debug/ShowMeTheOlly:
```
robbkidd, line 9: I haven't tested yet in a real collector whether the debug exporter is happy with being given a name.

TylerHelmuth: it can have a name.

pkg/config/components/OTelDebugExporter.yaml
TylerHelmuth: I think only surfacing `Verbosity` is good for now

pkg/config/components/OTelDebugExporter.yaml
TylerHelmuth: I don't think validations is implemented yet based on the usage of https://github.com/honeycombio/hpsf/blob/11c6bf27aaae7e6c8f9c7ca5e668cbb875bbf469/pkg/config/templateComponent.go#L45

pkg/translator/testdata/simple_grpc_hpsf.yaml
```diff
@@ -9,6 +9,8 @@ components:
       - name: Port
         value: 4317
         type: number
+  - name: ShowMeTheOlly
+    kind: OTelDebugExporter
```
codeboten, line 13: should this end up in a different test? its not just a grpc test anymore 😬

robbkidd: It probably should, yep. I started here for convenience, but I'll pop it out into a separate test case.

TylerHelmuth (approved):

pkg/config/components/OTelDebugExporter.yaml
robbkidd: I have changed this to `validations: [nonblank]` like other string-value properties on existing components. I suppose we've got a latent feature here for an `enum` validation or something.

PR #21, opened 2025-01-24
feat: add support for configuring OTLP headers

codeboten:

- This adds support for setting a map of values in addition to string values. Note that headers is currently nullable as well, we may want to add a way specify whether a property is nullable or not.

- added headers property + data template to grpc/http exporter
- adding a mechanism to `applyTemplate` to support types other than strings to be returned

### DIFF
--- a/pkg/config/components/OTelGRPCExporter.yaml
+++ b/pkg/config/components/OTelGRPCExporter.yaml
@@ -35,8 +35,8 @@ properties:
       Sending data to a backend may require additional headers to be
       configured. This properties supports sending a map of header keys and
       values.
-    type: kvlist
-    validations: [kvlist]
+    type: map
+    validations: [map]
 templates:
   - kind: collector_config
     name: otel_grpc_exporter_collector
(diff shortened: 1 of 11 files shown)


pkg/config/components/OTelGRPCExporter.yaml
```diff
@@ -40,4 +48,5 @@ templates:
     data:
       - key: "{{ .ComponentName }}.endpoint"
         value: "{{ firstNonblank .HProps.Host .User.Host .Props.Host.Default }}:{{ firstNonblank .HProps.Port .User.Port .Props.Port.Default }}"
-      # service is not part of the template, it's generated automatically by the collectorConfig
\ No newline at end of file
+      - key: "{{ .ComponentName }}.headers"
+        value: "{{ .HProps.Headers }}"
```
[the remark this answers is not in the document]
codeboten: Done

pkg/config/templateComponent.go
```diff
@@ -177,6 +178,32 @@ func (t *TemplateComponent) applyTemplate(tmplText string, userdata map[string]a
...
+	case string:
+		if tmplVal == "" || !strings.Contains(k, "{{") {
+			return k, nil
+		}
+		// the content of the template is a single value
+		// if we're just including a specific value, don't expand it
+		// as a string, instead return the value
+		// FIX: this is a hacky way of returning the value rather than
+		//      expanding the template
+		re := regexp.MustCompile(`{{ \.HProps\.(.*) }}`)
```
TylerHelmuth, line 192: IDK if it matters, but should we make this re a var/const since it doesn't change?

TylerHelmuth (approved):

pkg/config/templateComponent.go
codeboten: Done

PR #23, opened 2025-02-03
fix: DeterministicSampler component template update

robbkidd:

The default HPSF configuration wouldn't work.

- Added a test for the Refinery rules output from the default config
    - Default config uses the Refinery DeterministicSampler component and that component's template syntax hadn't kept up with recent changes.
- Brought the DeterministicSampler component up-to-date to use a different syntax for the key's dotted representation and to use firstNonblank and HProps for a value given from the input HPSF document


kentquirk (approved):
I like the solution.

MikeGoldsmith (approved):

robbkidd:
Merging! Will update other component templates in other PRs.

PR #25, opened 2025-02-05
fix: handle non-string properties in components and update them

robbkidd:
https://app.asana.com/0/1208469935858869/1209310061786981

- Syntax has changed some for component templates since some were written. 
- HPSF was coercing everything to strings, which caused problems. We really wanted to handle normal datatypes.

- Update template syntax
- Add tests
- Allow all (most?) of the places to have actual data types and allow the templates to express them. We might be able to do still better, but what we have here now works pretty well.
- Fix the templates and tests to do the right things

### DIFF
--- a/pkg/config/helpers.go
+++ b/pkg/config/helpers.go
@@ -9,29 +9,47 @@ import (
 	"time"
 )
 
+// We use GroupSeparator, FieldSeparator, and RecordSeparator as delimiters so
+// that we are unlikely to see a false positive with user data. We probably
+// could have done this with a more sophisticated encoding/decoding but we don't
+// think it matters in the context of templated configuration files.
 const (
+	GroupSeparator  = "\x1d" // ASCII code 29, the group separator
 	RecordSeparator = "\x1e" // ASCII code 30, the record separator
 	FieldSeparator  = "\x1f" // ASCII code 31, the unit (field) separator
+)
+
+// Prefixes we support:
+const (
+	IntPrefix   = "int" + GroupSeparator
+	BoolPrefix  = "bool" + GroupSeparator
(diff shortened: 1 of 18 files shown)


kentquirk:
I hope you don't mind a friendly commit. I guess I could have done it as a PR but as this is still a draft, I figured it would be OK.

What I've done is change it so that if a template emits something that looks like a list: `[ x, y, z ]`, it will take it apart and store it as a slice. This means the YAML renderer will do the right thing. 

I also changed the name and semantics of firstNonblank to firstNonZero, and made it so that if it's given a list, it uses the syntax above. I think this means that firstNonZero behaves the way we want it to even if the objects are lists instead of plain data. 

But also, if we just generate a template that has square brackets around it, it will be emitted as a list when rendered to a config.

kentquirk:
I just added a further improvement. Right now, you can add decorations to the templates (see the emasampler) to coerce types. But maybe we can do this automatically with the property types specified in the components.

TylerHelmuth (approved):

kentquirk (approved):
I'm now going to officially approve the parts of this I didn't write.

pkg/config/helpers.go
```diff
@@ -1,13 +1,19 @@
...
 	"fmt"
 	"regexp"
 	"strings"
 	"text/template"
 	"time"
 )
 
+const (
+	RecordSeparator = "\x1e" // ASCII code 30, the record separator
+	FieldSeparator  = "\x1f" // ASCII code 31, the unit (field) separator
```
robbkidd, line 14: emphatic nodding yes

kentquirk:
OK, I'm happy enough with this and it's a lot better than it was. Gonna merge and move on.

PR #26, opened 2025-02-17
fix: Fix encoding of map to handle nesting

kentquirk:

- Nested maps were getting mangled by the map encoder. 
- This was caught in tests -- except that that test wasn't running on my VS Code test runner OR in branch testing for some reason, so the failure leaked into main. It's working in both places now. I don't know why it failed.

- Change map->string encoding from simple and hand-coded to a JSON encoding.


MikeGoldsmith (approved):
We're just eating errors from Encode and Unmarshal. It doesn't look like we have a way to deal with errors right now though.

Should we create a follow-up issue to track that for the future?

PR #27, opened 2025-02-18
feat: First 2 templates for configurations

kentquirk:

- This provides the content of 2 initial template configurations for use in the UI. These will end up being moved to a database eventually.
- We're planning to also have a template for sending things to cold storage (and eventually many more) but we don't have the components to support that yet.

- Proxy template
- EMA Throughput template
- TraceConverter component improved

### DIFF
--- a/pkg/config/templates/emathroughput.yaml
+++ b/pkg/config/templates/emathroughput.yaml
@@ -1,0 +1,58 @@
+name: EMA Throughput Sampling
+description: |
+  Limit the throughput of traces to Honeycomb using an EMA sampler. This template
+  accepts traces from an OTel receiver, converts them to Honeycomb format, and
+  then samples them using an EMA throughput sampler before exporting them to
+  Honeycomb. Control the volume of traces by modifying the GoalThroughput value in
+  the properties section of the EMAThroughputSampler component.
+components:
+  - name: TraceGRPC_1
+    kind: TraceGRPC
+  - name: TraceConverter_1
+    kind: TraceConverter
+  - name: EMAThroughputSampler_1
+    kind: EMAThroughputSampler
+  - name: HoneycombExporter_1
+    kind: HoneycombExporter
+connections:
+  - source:
+      component: TraceGRPC_1
+      port: TraceOut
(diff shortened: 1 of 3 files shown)


pkg/config/templates/emathroughput.yaml
```diff
@@ -0,0 +1,58 @@
+name: EMA Throughput Sampling
+description: |
+  Limit the throughput of traces to Honeycomb using an EMA sampler. This template
+  accepts traces from an OTel receiver, converts them to Honeycomb format, and
+  then samples them using an EMA throughput sampler before exporting them to
+  Honeycomb. Control the volume of traces by modifying the GoalThroughput value in
+  the properties section of the EMAThroughputSampler component.
+components:
+  - name: TraceGRPC_1
+    kind: TraceGRPC
```
JamieDanielson, line 10: I am a little confused about the relationship between components here and the pkg/config/components. Should the "kind" match between them? Or am I misunderstanding how these are defined?

kentquirk: You're right -- I just figured this out myself when doing some stuff for Sarah. I'll clean it up.

kentquirk: OK, I've updated, and tested that the given templates generate valid configs and rules.

JamieDanielson (commented):
The proxy template looks good. There's an issue with the emathroughput template that I believe is because of an issue in TraceConverter. The Collector Config piece has an extra protocol section in Exporters that is not valid. I'm still working through that but I wonder if we separate that into its own PR to merge in while we work through the TraceConverter.

kentquirk:
Do we just need to delete ".protocols" from one of the data keys?

kentquirk:
Turns out it *was* just the TraceConverter component. I've fixed it there.

JamieDanielson (approved):
Looks great!
Question not necessarily needed in this PR: do we want to consider default fields for the FieldList in the EMA Sampler? We suggest HTTP method and status code, perhaps we could do that so it's usable out of the box... A separate conversation!

PR #28, opened 2025-02-20
feat: refactor translator so we can inject components from a database

kentquirk:

- HPSF is currently loading its components from an embedded directory. We want to be able to load them from a database.

- Create a NewEmptyTranslator function
- Deprecate but do not remove NewTranslator (so as not to break Hound)
- Create functions to load and inject components
- Call the new functions from main and fix tests

This PR depends on #27 and should be merged after that one.


MikeGoldsmith (approved):
Looks good. This might be a good way to synchronise changes to these templates automatically.

PR #29, opened 2025-02-21
maint: add make targets for validating configs

JamieDanielson:

- easy way to generate configs from templates
- easy way to validate configs

- Add make targets for `validate` and `test_template` 
- Generated configs from templates get placed in tmp/ directory that is gitignored

Example usage

```sh
make test_template TEMPLATE=pkg/config/templates/proxy
make validate CONFIG=examples/hpsf2.yaml
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
 .DS_Store
 .vscode
 test_results/
+tmp/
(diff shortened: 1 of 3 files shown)


JamieDanielson:
I made the template PR the base since this adds testing for templates but I guess this could also be added separately 🤷

robbkidd (commented):
Yes! ... HA HA HA ... YES!

TylerHelmuth (approved):

PR #30, opened 2025-02-25
fix: add a step to validate all examples in CI

codeboten:
This will ensure that changes made to the code will work with the examples listed in the examples directory.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -44,6 +44,9 @@ validate:
 	@echo "+++ validating config $(CONFIG)"
 	@echo
 	go run ./cmd/hpsf -i $(CONFIG) validate
+	for format in rConfig rRules cConfig ; do \
+		go run ./cmd/hpsf -i $(CONFIG) $${format} || exit 1; \
+	done
 
 .PHONY: validate_all
 validate_all: examples/*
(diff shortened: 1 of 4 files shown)


Makefile
```diff
@@ -44,3 +44,12 @@ validate:
 	@echo "+++ validating config $(CONFIG)"
 	@echo
 	go run ./cmd/hpsf -i $(CONFIG) validate
+	for format in rConfig rRules cConfig ; do \
+		go run ./cmd/hpsf -i $(CONFIG) $${format} || exit 1; \
+	done
```
JamieDanielson, line 49:
```suggestion
	@echo
	@echo "+++ generating configs for $(CONFIG)"
	@echo
	for format in rConfig rRules cConfig ; do \
		echo "\n+++ generating config for $${format}\n"; \
		go run ./cmd/hpsf -i $(CONFIG) $${format} || exit 1; \
	done
```
I love this! Wonder if we can add some extra "helper text" line to convey what each piece is doing.

JamieDanielson (approved):
Thank you for doing this! Added a suggestion to make it even more clarifying, feel free to ignore or update as needed!

Makefile
codeboten: added some text around this 👍🏻

PR #31, opened 2025-02-26
maint: rearrange components to a new place

kentquirk:

- We've decided to use this repository as the source of truth for HPSF components and templates, at least for the time being. We needed to move them from embedded within config out into their own package.
- By putting them into a separate package with the HPSF namespace, we can embed them into the filesystem, which makes it possible to wrap the filesystem into some helper functions.

- Move the components into `pkg/data/components` 
- Move the templates into `pkg/data/templates`
- Embed those filesystems into the data package
- Move and rename `LoadEmbeddedComponents` into data
- Update existing uses and tests
- Normalize and correct the fields
- Add version to component struct and components
- Add loader for templates


pkg/data/templates/forSarah.yaml
```diff
@@ -0,0 +1,66 @@
+components:
```
codeboten, line 1: can we rename this file to what its representing

kentquirk: it was accidentally committed. I'll delete it.

pkg/data/components/OTelDebugExporter.yaml
```diff
@@ -1,5 +1,6 @@
-name: otel_debug_exporter
 kind: OTelDebugExporter
+name: otel_debug_exporter
```
JamieDanielson, line 2: The name on some of the other components use some camel-casing e.g. `EMA_Throughput_Sampler`. Is there a way we can make these all consistent? Also happy to do as a follow-up once it's decided. Main idea of consistent casing style would be because it shows up in the UI.

JamieDanielson (commented):
With a quick run-through it looks good, haven't had time to dig too deeply. Left one suggestion on naming which doesn't have to be blocking for this.

pkg/data/components/OTelDebugExporter.yaml
kentquirk: I decided that we can use actually natural naming including spaces, so I renamed them all to look like `OTel Receiver`

JamieDanielson (approved):

PR #33, opened 2025-02-27
maint: add proxy example and test for collector config

JamieDanielson:

- internal asana task https://app.asana.com/0/1208469935858869/1209503664352030

- add example that demonstrates basic property overrides for proxy template (otlp collector) 
- add Makefile target for smoke and unsmoke that simply generates the collector config from the proxy template and tries to run it in Docker. This isn't in CI yet but can be run locally.

**Note**: When this is generated it uses the "Name" field in the collector config. If the name has spaces it will be invalid in Collector config. We'll want to normalize that but maybe as a follow-up. We can use this example for testing.

For example, `otlp/MyCompany OTel Receiver` should become something like `otlp/MyCompany_OTel_Receiver`

This can be confirmed with a quick run in docker (basic "smoke" test added, name tbd):

```sh
make smoke
make unsmoke
```

results in collector logs if spaces are in name (truncated for brevity)

```sh
Error: failed to get config: cannot unmarshal the configuration: decoding failed due to the following error(s):

error decoding 'receivers': in "MyCompany OTel Receiver" id: invalid character(s) in name "MyCompany OTel Receiver"
error decoding 'exporters': in "MyCompany HTTP Exporter" id: invalid character(s) in name "MyCompany HTTP Exporter"
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -65,7 +65,7 @@ smoke:
 	@echo
 	mkdir -p tmp
 	go run ./cmd/hpsf -i ./examples/hpsfProxy.yaml -o tmp/hpsfProxy.cconfig.yaml cConfig
-	docker run --rm -d \
+	docker run --rm -it \
 		--name smoke-proxy \
     -p 4227-4228:4227-4228 \
 		-v ./tmp/hpsfProxy.cconfig.yaml:/etc/otelcol/config.yaml \
--- a/Makefile
+++ b/Makefile
@@ -56,3 +56,24 @@ validate_all: examples/*
 	for file in $^ ; do \
 		$(MAKE) validate CONFIG=$${file} || exit 1; \
 	done
+
+.PHONY: smoke
+smoke:
+	@echo
+	@echo "+++ basic smoke test - does it start?"
+	@echo "+++ if so, make unsmoke after this"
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -56,3 +56,24 @@ validate_all: examples/*
...
+
+.PHONY: smoke
+smoke:
+	@echo
+	@echo "+++ basic smoke test - does it start?"
+	@echo "+++ if so, make unsmoke after this"
+	@echo
+	mkdir -p tmp
+	go run ./cmd/hpsf -i ./examples/hpsfProxy.yaml -o tmp/hpsfProxy.cconfig.yaml cConfig
+	docker run --rm -d \
```
MikeGoldsmith, line 68: If we use `-it` instead of `-d`, the smoke test will run the collector and connect it's output directly to the terminal. We then don't need to `unsmoke` to tear down the container either as it'll be cleaned once you ctrl-c to exit from `smoke`.

```suggestion
	docker run --rm -it \
```

JamieDanielson: Sure that works. I'd like to run it in CI eventually but this works for local!

codeboten (approved):

PR #34, opened 2025-02-28
feat: automatically inject usage components to collector render

TylerHelmuth:

https://app.asana.com/0/1208469935858869/1209459312125798

- update collector render function to inject some required honeycomb components.


pkg/translator/testdata/default_collector_config.yaml
```diff
@@ -1,2 +1,9 @@
+processors:
+    usage:
+        honeycombextensionID: honeycomb
+extensions:
+    honeycomb:
+        opampextensionID: opamp
```
TylerHelmuth, line 6: A downside to this approach is that it produces a collector config that is dependent on the opampsupervisor adding an opampextension with the ID `opamp`.

Currently the opampsupervisor ALWAYS adds its own opampextension and it will replace any opampextension that the provided config already has. It always uses the ID `opamp` right now.

But it is possible the opampsupervisor changes in the future and that results in a (I think?) breaking change for HPSF from something outside our control.

kentquirk:
Interesting. This is pretty good. I need to look at it again tomorrow. I think it *might* be cleaner to inject stuff into NewCollectorConfig as pre-existing Sections but it's late for me so i'm not going to try.

pkg/translator/testdata/default_collector_config.yaml
MikeGoldsmith: We have discussed recording compatibility version numbers on components before, eg works with Collector v0.120.0+. That could help with knowing what extension ID to use or other config for things outside of our control. 

If a pipeline had a target collector version, we could select the appropriate usage & extension component version with valid config.

kentquirk:
OK, I looked at it in the cold light of day. In general, this seems like the way to do it; I tried to put it in NewCollectorConfig but that doesn't work since not all signals are used. 

The one thing I'd like to do, if you don't mind, is pull out the custom modifications from `RenderYAML` and put them into a helper method on the `collectorConfigFormat` type. I'd like to keep the render method cleaner if possible.

pkg/config/tmpl/collectorconfig.go
```diff
@@ -160,3 +162,32 @@ func NewCollectorConfig() *CollectorConfig {
 	cc := CollectorConfig{Sections: make(map[string]DottedConfig)}
 	return &cc
 }
+
+// injectHoneycombUsageComponents ensures the collector configuration always has the necessary honeycomb
+// components for measuring usage.
+func injectHoneycombUsageComponents(f collectorConfigFormat) collectorConfigFormat {
```
kentquirk, line 168: I was thinking this would be
```suggestion
func (f *collectorConfigFormat) injectHoneycombUsageComponents() {
```
mainly because I prefer member functions to free functions.

kentquirk (approved):
One suggestion, but it's not enough to block it.

TylerHelmuth:
Depends on https://github.com/honeycombio/opentelemetry-collector-configs/pull/148

kentquirk (approved):
Re-reviewed, still approve.

pkg/config/tmpl/collectorconfig_test.go
```diff
@@ -10,25 +10,32 @@ func TestCollectorConfig_RenderYAML(t *testing.T) {
...
+    usage: {}
+extensions:
+    honeycomb: {}
 service:
+    extensions: [honeycomb]
     pipelines:
         traces:
             receivers: [otlp]
-            processors: []
+            processors: [usage, batch]
```
codeboten, line 29: not seeing the batch processor defined here, this would result in an error

codeboten (approved):
non blocking test comment

PR #35, opened 2025-03-03
fix: Improvements to validation

kentquirk:

- Cleaning up the validation logic

- Fix up property types (number -> int and float, add map)
- Adjust property definitions in various places
- Add validation for names of top-level keys in HPSF files
- Adjust makefile to check templates as well as examples

## Still to do

- [ ] Add validation of template components
- [ ] Add makefile steps to validate them
- [ ] Add validation of HPSF property values including their validation steps specified in the components

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -52,7 +52,7 @@ validate:
 	done
 
 .PHONY: validate_all
-validate_all: examples/*
+validate_all: examples/hpsf* pkg/data/templates/*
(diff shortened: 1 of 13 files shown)


JamieDanielson:
Doing another quick review; we've decided we'll put those 3 "Still TODO" items in a follow-up PR so these changes can get into Main.

JamieDanielson (approved):

PR #36, opened 2025-03-04
feat: Add style, tags, and status to TemplateComponents

kentquirk:

- Since this is now the source of truth for components, we need all the fields the database would have. 

- Adds Tags, Style and Status to the component data structure
- Ensures that Style and Status are lowercase in YAML but uppercase in the data by defining types for those fields
- Updates all the components to have appropriate values of the new fields

### DIFF
--- a/pkg/config/templateComponent.go
+++ b/pkg/config/templateComponent.go
@@ -11,6 +11,7 @@ import (
 
 	"github.com/honeycombio/hpsf/pkg/config/tmpl"
 	"github.com/honeycombio/hpsf/pkg/hpsf"
+	y "gopkg.in/yaml.v3"
 )
 
 // This is the Go support for components read as data.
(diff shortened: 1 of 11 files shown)


pkg/config/templateComponent.go
```diff
@@ -64,6 +65,73 @@ type TemplateData struct {
 	Data   []any
 }
 
+type ComponentStyle string
```
MikeGoldsmith, line 68: ComponentType is the way the components were grouped into base, meta and template components.

Style is how we want the component to be styled in the UI. Style is a freeform text value right now, but could be made into an number like component group?

For example:
- receiver
- processor
- sampler
- exporter
- etc

pkg/config/templateComponent.go
```diff
@@ -314,3 +388,12 @@ func (t *TemplateComponent) generateCollectorConfig(ct collectorTemplate, userda
 	}
 	return config, nil
 }
+
+func (t *TemplateComponent) AsYAML() ([]byte, error) {
```
MikeGoldsmith, line 392: Do we want this to return `[]byte` or a `string`?

pkg/data/components/FilterProcessor.yaml
```diff
@@ -1,10 +1,19 @@
 kind: FilterProcessor
 name: Filter Processor
+style: base
+status: development
 version: v0.1.0
 summary: Processor that can be used to filter telemetry.
 description: |
   Filters traces, metrics, and logs based on rules defined
   in the configuration.
+tags:
```
MikeGoldsmith, line 10: 🎉

JamieDanielson (approved):
Looks good and it seems all of Mike's requests were addressed. I hope I didn't cause too much issue with my other PR for default nop pipelines. I meant to base it on this branch 😬

kentquirk:
I see no conflicts, but just to be sure I'll merge main now.

PR #37, opened 2025-03-05
fix: deterministic sampler should use SampleRate

JamieDanielson:

- Deterministic Sampler should use SampleRate, not GoalThroughput

- Update DeterministicSampler component to look at SampleRate
- Update expected default refinery rules


kentquirk (approved):
💯

PR #38, opened 2025-03-05
fix: add nop pipeline for valid collector config used for initial setup

JamieDanielson:

- Default nop pipeline for valid collector config on initial startup
- fixes [internal asana task](https://app.asana.com/0/1208469935858869/1209503664352023) 

-  add nop receiver and nop exporter components
- add `case map[string]string` for template validation, because nop receiver/exporter should be valid with empty brace
- update `default_collector_config` testdata to show expected valid config with nop pipeline

Note this doesn't actually generate collector config in the `validate` case that I would expect. `make validate` runs against `examples/hpsf.yaml` and still shows empty pipelines.

Note: Some of this was edited since opened bc it appears caching made some things act up, so I've removed the extra unnecessary changes 😄


pkg/config/templateComponent.go
```diff
@@ -122,6 +122,9 @@ func (t *TemplateComponent) Props() map[string]TemplateProperty {
 }
 
 func (t *TemplateComponent) ComponentName() string {
+	if t.CollName == "nop" {
+		return t.CollName + "/" + t.Name
```
JamieDanielson, line 126: The `t.hpsf.GetSafeName()` was panicking on the Nop components, so I special-cased it for now as a time-boxed effort.

pkg/data/components/NopExporter.yaml
```diff
@@ -0,0 +1,22 @@
+kind: NopExporter
+name: DefaultNopExporter
+version: v0.1.0
+summary: Nop exporter
+description: |
+  Simple nop exporter.
+  This exporter does nothing, is required for minimal collector.
+ports:
+  - name: Traces
```
codeboten, line 9: i could see this being useful for additional signals, when someone wants to play around with configuration of their pipelines but dont want to emit anything just yet

codeboten (approved):

pkg/data/components/NopExporter.yaml
JamieDanielson: Agreed, I think we can do more with it. Right now it's basically the smallest possible component 😄

PR #39, opened 2025-03-07
feat: add insecure config to OTel exporters

codeboten:

- allow users to send data w/o TLS

- adding a property to otel exporters for insecure

### DIFF
--- a/pkg/data/components/OTelGRPCExporter.yaml
+++ b/pkg/data/components/OTelGRPCExporter.yaml
@@ -47,6 +47,12 @@ properties:
       values.
     type: map
     validations: []
+  - name: Insecure
+    summary: Provide a way to disable TLS export.
+    description: |
+      Can be used to send data without TLS.
+    type: bool
+    validations: [bool]
 templates:
   - kind: collector_config
     name: otel_grpc_exporter_collector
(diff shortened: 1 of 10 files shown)


kentquirk (commented):
You'll need to update the templates to use this property or it's not going to do anything.

pkg/data/components/OTelGRPCExporter.yaml
```diff
@@ -58,6 +65,8 @@ templates:
     data:
       - key: "{{ .ComponentName }}.endpoint"
         value: "{{ firstNonZero .HProps.Host .User.Host .Props.Host.Default }}:{{ firstNonZero .HProps.Port .User.Port .Props.Port.Default }}"
+      - key: "{{ .ComponentName }}.tls.insecure"
+        value: {{ firstNonZero .HProps.Insecure .User.Insecure .Props.Insecure.Default }}
```
kentquirk, line 69: Suggestion: use suppress_if to only include this if the result is true.

codeboten: good call, used suppress and added a test for insecure

JamieDanielson (approved):

PR #40, opened 2025-03-11
feat: add comment property, use advanced property

kentquirk:

- TemplateComponents now have a Comment property
- Several of the common components now have some of their properties marked with the Advanced flag

The advanced flag is an indication that most configurations will not need to modify that property, so it is hidden by default in the UI under a control that looks like "> Advanced". Clicking that control will reveal all those properties. It's intended to allow components to have a lot of configurability but to hide it from the average user.

- Add comment field to TemplateComponent
- delete unused function that was mistakenly committed
- Mark some properties as Advanced

### DIFF
--- a/pkg/data/components/OTelGRPCExporter.yaml
+++ b/pkg/data/components/OTelGRPCExporter.yaml
@@ -24,21 +24,6 @@ ports:
     direction: input
     type: OTelLogs
 properties:
-  - name: Host
-    summary: The hostname or IP address to send data to
-    description: |
-      Hostname or IP address on which to send outgoing GRPC traffic.
-    type: string
-    validations: [nonblank]
-    default: https://api.honeycomb.io
-  - name: Port
-    summary: The port on which to send gRPC traffic.
-    description: |
-      The port on which to send outgoing gRPC traffic. Default is 443, which is
-      the value expected by Honeycomb. The OTel standard for gRPC is 4317.
-    type: int
-    validations: []
-    default: 443
   - name: Headers
     summary: Headers to emit when sending gRPC traffic.
     description: |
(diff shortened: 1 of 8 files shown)


pkg/data/components/OTelGRPCExporter.yaml
```diff
@@ -24,13 +24,22 @@ ports:
...
+    type: map
+    validations: []
   - name: Host
     summary: The hostname or IP address to send data to
     description: |
       Hostname or IP address on which to send outgoing GRPC traffic.
     type: string
     validations: [nonblank]
     default: https://api.honeycomb.io
+    advanced: true
```
codeboten, line 42: what does it mean for a property to be advanced?

kentquirk: I updated the PR description above to include that information.

pkg/data/components/OTelGRPCExporter.yaml
```diff
@@ -24,13 +24,22 @@ ports:
...
     type: OTelLogs
 properties:
+  - name: Headers
+    summary: Headers to emit when sending gRPC traffic.
+    description: |
+      Sending data to a backend may require additional headers to be
+      configured. This property supports sending a map of header keys and
+      values.
+    type: map
+    validations: []
```
JamieDanielson, line 34: Oh good idea, assume this was moved to the top bc it's not advanced, and the advanced properties will be separated below it.

pkg/data/components/OTelHTTPExporter.yaml
```diff
@@ -39,21 +48,15 @@ properties:
...
     default: false
-  - name: Headers
-    summary: Headers to emit when sending HTTP traffic.
-    description: |
-      Sending data to a backend may require additional headers to be
-      configured. This properties supports sending a map of header keys and
-      values.
-    type: map
-    validations: []
+    advanced: true
```
JamieDanielson, line 59: Later we'll want to think about whether insecure is advanced, or if we should make any automatic decisions or add additional context. We found that insecure true was necessary when sending to refinery, or another collector, but insecure false is necessary for sending directly to honeycomb.

JamieDanielson (approved):

pkg/data/components/OTelHTTPExporter.yaml
kentquirk: "Send this to Refinery" is a separate component -- that's the TraceConverter.

PR #42, opened 2025-03-19
feat: new error types

kentquirk:

- The validation routine as written returned []error. That's both nonidiomatic and also awkward.
- This creates a new hpsf.Error type that can be either a warning or an error, and contains location detail
- And also revises the validation.Error type to validation.Result -- which is a collection of errors 
- And then uses them; we have no warnings yet but we'll get there
- Minor tweaks to validation code
- Add a test for failing validations to make sure the system works as advertised

This PR will require changes to hound once it merges -- there are breaking changes to the validate functions. I'll get to those next.

### DIFF
--- a/cmd/hpsf/main.go
+++ b/cmd/hpsf/main.go
@@ -143,11 +143,20 @@ func main() {
 		}
 
 		// validate the HPSF
-		errors := hpsf.Validate()
-		if len(errors) > 0 {
-			for _, e := range errors {
-				log.Printf("error: %v", e)
+		verrors := hpsf.Validate()
+		if verrors == nil {
+			log.Printf("HPSF is valid")
+			os.Exit(0)
+		}
+
+		if hErr, ok := verrors.(validator.Error); ok {
+			log.Printf("error: %v", hErr.Msg)
+			for _, e := range hErr.Details {
+				log.Printf("  error: %v", e)
 			}
+			os.Exit(1)
+		} else {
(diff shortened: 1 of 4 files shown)


cmd/hpsf/main.go
```diff
@@ -143,12 +143,21 @@ func main() {
...
-		if len(errors) > 0 {
-			for _, e := range errors {
-				log.Printf("error: %v", e)
+		verrors := hpsf.Validate()
+		if verrors == nil {
+			log.Printf("HPSF is valid")
+			os.Exit(0)
+		}
+
+		if hErr, ok := verrors.(validator.Error); ok {
```
MikeGoldsmith, line 152: The valid HPSF log entry after the if/else is now redundant. Can we move the os.Exit to after the if/else or change the verrors == nil to != and use the os.Exit inside that branch?

pkg/validator/validator.go
```diff
@@ -1,25 +1,65 @@
...
 	return e.Msg
 }
 
-func NewError(msg string) error {
-	return Error{Msg: msg}
+func (e Error) Unwrap() []error {
+	return e.Details
+}
+
+func NewError(msg string) Error {
```
MikeGoldsmith, line 20: It kinda feels weird we're calling NewError which may not be an error, but a grouping of errors which we check the length of later in ErrOrNil. Would renaming the struct to Validator or similar help?

MikeGoldsmith (approved):

PR #45, opened 2025-03-20
feat: add metrics/logs to nop components, all signals to debug exporter

codeboten:

- Ensures nop receiver/exporter and for the debug exporter can be connected for all signals.

- This adds all the signals for the nop receiver/exporter and for the debug exporter.


kentquirk (approved):
I *almost* added this last night! Thanks for catching it.

PR #46, opened 2025-03-20
bugfix: trace converter must support one of http or grpc

codeboten:

- The TraceConverter was producing invalid configuration.

- Updated the template to only use HTTP for export


MikeGoldsmith (approved):
Looks good. Should we add a gRPC equivalent component?

PR #47, opened 2025-03-21
feat: add headers/insecure support for TraceConverter

codeboten:
These fields are needded to be able to configure the exporter w/ all the right fields.

### DIFF
--- a/pkg/data/components/TraceConverter.yaml
+++ b/pkg/data/components/TraceConverter.yaml
@@ -36,7 +36,7 @@ properties:
     validations: [nonblank]
     default: ${STRAWS_REFINERY_POD_IP}
     advanced: true
-  - name: HTTPPort
+  - name: Port
     summary: The port on which Refinery is listening for HTTP traffic.
     description: |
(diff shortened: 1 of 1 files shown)


pkg/data/components/TraceConverter.yaml
```diff
@@ -45,6 +45,22 @@ properties:
...
     default: 4318
     advanced: true
+  - name: Headers
+    summary: Headers to emit when sending HTTP traffic.
+    description: |
+      Sending data to a backend may require additional headers to be
+      configured. This properties supports sending a map of header keys and
+      values.
+    type: map
+    validations: []
```
kentquirk, line 55: I think this should be an advanced property (in this component) because under normal use, they won't need to add anything to send the data to Refinery.

kentquirk (approved):
One note, but I can live with it either way.

pkg/data/components/TraceConverter.yaml
codeboten: think that will depend on whether or not a send key is configured on the honeycomb exporter. will merge like this for now and revised if needed next week

PR #48, opened 2025-03-25
feat: add batch processor automatically

codeboten:

- Batching should be part of the pipeline by default.

- Automatically append the batch processor to the generated list of processors. This will be changed in the near future


MikeGoldsmith (approved):
Looks good, just a couple of small changes for replacing usage with batch

PR #52, opened 2025-03-27
feat: Update the default template to be a complete embedded template

MikeGoldsmith:

Updates the default HPSF config to use the embedded proxy template.

- Add new default template (copy of proxy)
- Add LoadEmbeddedDefaultTemplate and DefaultConfigurationKind that can be used to retrieve the default template
- Move default config test from hspf to data package
- Add AsYAML func to hspf type (same as we use for TemplateComponent)
- Update translator tests for expected default config to match proxy template output

### DIFF
--- a/pkg/data/loadEmbedded.go
+++ b/pkg/data/loadEmbedded.go
@@ -11,6 +11,8 @@ import (
 	y "gopkg.in/yaml.v3"
 )
 
+const DefaultConfigurationKind = "TemplateProxy"
+
 // Reads a set of components from the local embedded filesystem (in the source, this is the
 // data/components directory) and loads them into a map of TemplateComponent by name.
 func LoadEmbeddedComponents() (map[string]config.TemplateComponent, error) {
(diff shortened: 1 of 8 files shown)


kentquirk:
I like the idea in general, but I think we'll probably want to evolve the proxy and the default separately. How about this:
* The default template name becomes `TemplateDefault`
* Make a copy of the current Proxy and name it the default

MikeGoldsmith:
> I like the idea in general, but I think we'll probably want to evolve the proxy and the default separately. How about this:
> 
> * The default template name becomes `TemplateDefault`
> * Make a copy of the current Proxy and name it the default

Added a new default template in [9d1716e](https://github.com/honeycombio/hpsf/pull/52/commits/9d1716e888ef6b450804af9d5e61538685dda5af)

JamieDanielson:
Do we still need a default refinery component so that starts up in the default install too?

kentquirk:
Yes, the default refinery should have a default deterministic sampler with a rate of 1. I thought we already had that.

MikeGoldsmith:
The defautl doesn't have any sampler. I'll add it to both default and proxy so they can generate valid configs.

The HPSF validator should have a rule that a sampler is required too then.

kentquirk:
OK. What I meant is that I thought that I had implemented it so that the Refinery Rules config generator inserted a default sampler if you didn't specify one. Did that part get deleted?

MikeGoldsmith:
I guess so - the Translator returns an empty refinery rules config if given HPSF that doesn't contain a sampler component.

Maybe we file a bug with the translator to always return a valid (runnable) rules definition even when no sampler component is specified?

MikeGoldsmith:
Filed bug. I'll look to fix that before merging this change otherwise we could leak invalid configurations to users of this package.

MikeGoldsmith:
Merge the fix from #54. This is ready for review now.

codeboten (approved):
This looks ok!

kentquirk (approved):
Yup.

PR #54, opened 2025-03-27
fix: Translator generates valid refinery rules when no sampler component is provided

MikeGoldsmith:

When generating configs, if no sampler component is provided the translator does not produce a valid refinery rules configuration (no default sampler).

This PR updates the sampler to check if a sampler was provided in the list of components, and if not, adds a default deterministic sampler with a sample rate of 1.

- Add style property to hpsf.Component
- Add `maybeAddDefaultSampler` that checks if there the components contains a sampler and adds one if not
- Add unit test


kentquirk (approved):
This looks pretty good! Might be all we need for now.

PR #60, opened 2025-03-30
feat: Validate properties with the property rules

kentquirk:

- Validation wasn't taking the validation rules into account. Now it does.

- Add IsEmpty method to Result
- Move TemplateProperty to its own file
- Add Validate method to the TemplateProperty and support for it
- Add ValidateConfig method to the translator -- it returns a Result which is a collection of errors.
- Implement a bunch of validation helpers
- Tweak the existing components to use them
- Fix up a bunch of testdata and examples that were based on old stuff


pkg/data/components/FilterProcessor.yaml
```diff
@@ -28,10 +28,10 @@ ports:
 properties:
   - name: Rules
     type: rule
-    validations: [rule]
+    validations: [filterrule]
   - name: Signal
     type: string
-    validations: [string] # TODO: validate a valid signal
+    validations: ["oneof(traces, metrics, logs)"] # TODO: validate a valid signal
```
MikeGoldsmith, line 34: We could add a custom validSignal rule that interally is made up using `oneof(traces, metrics, logs)`

pkg/translator/translator.go
```diff
@@ -73,17 +74,97 @@ func (t *Translator) MakeConfigComponent(component hpsf.Component) (config.Compo
...
+}
+
+// ValidateConfig validates the configuration of the HPSF document as it stands with respect to the
+// components and templates installed in the translator.
+// Note that it returns a validation.Result so that the errors can be collected and reported in a
+// structured way. This allows for multiple validation errors to be returned at once, rather than
+// stopping at the first error. This is useful for providing feedback to users on multiple issues
+// in their configuration.
+func (t *Translator) ValidateConfig(h *hpsf.HPSF) error {
+	if h == nil {
```
MikeGoldsmith, line 93: Is an empty (no components) HPSF document valid? If not, we should do a len check here too.

MikeGoldsmith (approved):
Looks great.

pkg/config/templateProperty.go
```diff
@@ -0,0 +1,305 @@
...
+	name := matches[1]
+	args := []string{}
+	// matches[2] contains the arguments
+	if len(matches) == 3 {
+		if len(matches[2]) > 0 {
+			args = argpat.Split(matches[2], -1) // split on commas, remove empty strings
+		}
+	}
+
+	switch name {
```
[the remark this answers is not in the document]
kentquirk: I'm of mixed feelings about that. I did consider it. In this case it's pretty harmless, but it sort of establishes a precedent. For example, should the arguments to oneof also be case insensitive? I think I prefer to not open that can of worms today.

pkg/data/components/FilterProcessor.yaml
kentquirk: yes; let's see if we think it's common enough to justify. The presence of HoneycombEvents complicates this a bit.

MikeGoldsmith: On review, I don't think we need this. There are going to be a mix of components with different support signals. Having an abstract one doesn't make it any clearer / easier to read.

pkg/translator/translator.go
kentquirk: I thought so, and I was wrong. At least it broke tests. So I reverted it.

PR #61, opened 2025-03-31
feat: Add refinery rules based sampler

MikeGoldsmith:

-

-

### DIFF
--- a/pkg/translator/translator_test.go
+++ b/pkg/translator/translator_test.go
@@ -13,7 +13,7 @@ import (
 	yamlv3 "gopkg.in/yaml.v3"
 )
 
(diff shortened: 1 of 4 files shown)


pkg/translator/translator_test.go
```diff
@@ -97,6 +97,57 @@ func TestGenerateConfig(t *testing.T) {
 	}
 }
 
+func TestGenerateRefineryRulesConfigs(t *testing.T) {
```
codeboten, line 100: not a blocker, but i would add the desired generated config type to the test struct and avoid the duplicated code (maybe include it in the test name to make it clear which config failed to generate

codeboten (approved):

PR #66, opened 2025-04-02
maint: Apply component property validations

MikeGoldsmith:

Component validations are not currently being applied. This update applies property validation and updates the components where necessary to conform to their validations.

- Update TemplateComponent's yaml name to so they are loaded into the component
- Use the property value, not type, when trying to coerce the value before applying a validation
- Update each components validations definition to use a yaml list instead of inline array which can cause issues when parsing strings
- Update HoneycombExporter testdata to set a valid URL
- Update Validation.Result's Unwrap to return a single error instead of an []error
  - The docs say []error should be okay, but it fails the [interface check](https://github.com/golang/go/blob/master/src/errors/wrap.go#L17) and always returned nil
- Remove filter processor's Rules `filterrule` as it doesn't exist yet
  - Opened #67 to track adding a custom validation
- Remove unused traceconverter collector config testdata files

### DIFF
--- a/pkg/data/components/LogDeduplicationProcessor.yaml
+++ b/pkg/data/components/LogDeduplicationProcessor.yaml
@@ -22,7 +22,8 @@ ports:
 properties:
   - name: Interval
     type: duration
-    validations: [duration]
+    validations:
+      - duration
     default: 60s
     advanced: true
   - name: CountAttribute
--- a/pkg/data/components/S3Exporter.yaml
+++ b/pkg/data/components/S3Exporter.yaml
@@ -30,7 +30,8 @@ properties:
     description: |
       The bucket in which to store the S3. This is a required field.
     type: string
-    validations: [noblanks]
+    validations:
+      - noblanks
   - name: Region
     summary: The region in which to store the S3.
     description: |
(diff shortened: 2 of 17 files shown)


kentquirk (commented):
Looks good. Thanks.

pkg/data/components/LogDeduplicationProcessor.yaml
```diff
@@ -22,7 +22,8 @@ ports:
 properties:
   - name: Interval
     type: duration
-    validations: [duration]
+    validations:
```
codeboten, line 25: this makes me wonder if we could leave validations empty and use the type for validation in such cases

pkg/data/components/S3Exporter.yaml
```diff
@@ -53,14 +54,18 @@ properties:
...
-    validations: [duration]
+    default: 5s
+    validations:
+      - duration
     advanced: true
   - name: Marshaler
     summary: The marshaler to use for the S3.
     description: |
       The marshaler to use for the S3. The default value is `otlp_json`.
     type: string
```
codeboten, line 65: any option that uses a oneof validation may benefit from a different type than string

kentquirk (approved):
Oops! Approved.

codeboten (approved):
🎉

pkg/data/components/LogDeduplicationProcessor.yaml
kentquirk: Type isn't always sufficient. Normal type validations, yes -- which is why there are no validations like "int" or "bool". But range, or limited numbers of strings need more flexibility

pkg/data/components/S3Exporter.yaml
MikeGoldsmith: Sure, I'll open a issue to track adding discrete types when using `oneof` validations.

PR #80, opened 2025-04-11
feat: set SendKeyMode when api key present

codeboten:

- This allows refinery to use the configured API key instead of the one present in the headers.

- added tests for presence of the API key and when it's not present
- ensure `suppress_if` is respected for all types of config
- updated honeycomb exporter with SendKeyMode template


kentquirk (commented):
Are you sure you need to add the nonempty helper? Go templating has javascript semantics, so my expectation is that this would work:

```
suppress_if: "{{ firstNonZero .HProps.APIKey .User.APIKey }}"
```

kentquirk (approved):
The helper doesn't actually do any harm, so I guess it makes sense to approve it either way.

PR #83, opened 2025-04-14
feat: Add smoke tests for refinery components

MikeGoldsmith:

Adds integration tests to verify each refinery component can generate config and rules files that a refinery docker image can start with.

- https://app.asana.com/1/72322038700732/project/1208469935858869/task/1209503669851437?focus=true

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -60,7 +60,7 @@ validate_all: examples/hpsf* pkg/data/templates/*
 CONFIG ?= examples/hpsf.yaml
 .PHONY: smoke_refinery
 smoke_refinery:
-	@echo generating refinery configs for component component $(CONFIG)
+	@echo generating refinery configs for component $(CONFIG)
 	mkdir -p tmp
 
 	# generate the configs from the provided file
@@ -76,10 +76,10 @@ smoke_refinery:
 
 	# check if the container is running
 	if [ "$$(docker inspect -f '{{.State.Running}}' 'smoke-refinery')" != "true" ]; then \
-		echo "+++ smoke test - container not running"; \
+		echo "+++ container not running"; \
 		exit 1; \
 	else \
-		echo "+++ smoke test - container running"; \
+		echo "+++ container is running"; \
 		docker kill 'smoke-refinery' > /dev/null; \
 	fi
 
--- a/Makefile
+++ b/Makefile
@@ -57,19 +57,39 @@ validate_all: examples/hpsf* pkg/data/templates/*
 		$(MAKE) validate CONFIG=$${file} || exit 1; \
 	done
 
-.PHONY: smoke
-smoke:
-	@echo
-	@echo "+++ basic smoke test - does it start?"
-	@echo "+++ if so, make unsmoke after this"
-	@echo
+CONFIG ?= examples/hpsf.yaml
+.PHONY: smoke_refinery
+smoke_refinery:
+	@echo generating refinery configs for component component $(CONFIG)
 	mkdir -p tmp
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -57,19 +57,39 @@ validate_all: examples/hpsf* pkg/data/templates/*
...
 	done
 
-.PHONY: smoke
-smoke:
-	@echo
-	@echo "+++ basic smoke test - does it start?"
-	@echo "+++ if so, make unsmoke after this"
-	@echo
+CONFIG ?= examples/hpsf.yaml
+.PHONY: smoke_refinery
```
codeboten, line 61: `make smoke_refinery` fails on its own

Makefile
```diff
@@ -57,19 +57,46 @@ validate_all: examples/hpsf* pkg/data/templates/*
...
 
-.PHONY: smoke
-smoke:
-	@echo
-	@echo "+++ basic smoke test - does it start?"
-	@echo "+++ if so, make unsmoke after this"
-	@echo
+.PHONY: .smoke_refinery
+#: run smoke test for refinery
+#: (usage: make smoke_refinery FILE=examples/hpsf.yaml)
```
codeboten, line 62:
```suggestion
```

codeboten (approved):

JamieDanielson (dismissed):
Looking good! Just need to update the smoke test as it doesn't fail on failed runs. For example the `filterprocessor_all` fails with an error:

2025/04/08 18:53:07 collector server run finished with error: invalid configuration: service::pipelines::traces: references processor "filter/drop_container_1" which is not configured

Also do we want to run this in CI? And/or do we need a way to distinguish between dev components (that are maybe expected to fail right now) and alpha/live components (that should pass)?

JamieDanielson:
> For example the `filterprocessor_all` fails with an error:

Actually I guess this should only be testing Refinery components, as the Collector smoke test is a separate ticket. So maybe that piece needs to be updated to only test Refinery files 🤔

MikeGoldsmith:
In my testing, only the refinery components are picked up:
```
make smoke
...
generating refinery configs for component pkg/data/components/DeterministicSampler.yaml
...
generating refinery configs for component pkg/data/components/EmaThroughput.yaml
...
generating refinery configs for component pkg/data/components/HoneycombExporter.yaml
```

MikeGoldsmith:
> In the initial ticket I was thinking we could take advantage of Refinery's built-in validation (`go run ./cmd/refinery -r rules.yaml -c config.yaml -V`) and instead of the bare components, use the full generated configs that test all the config changes (e.g. those in `pkg/translator/testdata/refinery_rules`, but I guess that could be a separate PR.

The new smoke test uses the official Refinery container and hpsf generates both the config and rules files to run with. I don't think running testdata file helps here. We could expand these smoke tests to verify templates later to make sure they are valid too.

> Otherwise the only other question is whether this should run in CI. I'll remove my request changes review.

Yeah, I agree these should be run in CI. I plan on doing the Collector one next so will tie into CI in that PR.

JamieDanielson:
🤔  I know this was merged already but I don't think this is testing what we think it is. If I set it to "sleep 100" instead of "sleep 1" and review the files (etc/refinery/rules.yaml) I only see the default sampler with a Sample Rate of 1, not the actual component. To confirm it was the default sampler, I updated the SampleRate in `translator.go` to SampleRate:123 and see that in the file in the container. Maybe the script isn't properly iterating through the components? 

etc/refinery/rules.yaml in docker container

PR #84, opened 2025-04-15
feat: Update Refinery smoke tests

MikeGoldsmith:

Updates smoke tests for generated refinery configs added in #83.

- closes internal issue https://github.com/honeycombio/pipeline-team/issues/259

- Update the make target to iterates HPSF templates in pkg/data/templates
- After generating configs, it starts docker containers with the generated configs and tests to see if they started successfully
- Update EMA throughput template to provide a default FieldList which is a require property

NOTE: This PR is blocked on the following PRs that fix some invalid components:
- https://github.com/honeycombio/hpsf/pull/86
- https://github.com/honeycombio/hpsf/pull/87


smoke_test.sh
```diff
@@ -0,0 +1,51 @@
...
+# use yq to remove the usage processor and honeycomb extension from collector config
+yq -i e \
+    'del(.processors.usage) | del(.extensions) | del(.service.extensions) | del(.service.pipelines.traces.processors[] | select(. == "usage"))' \
+    tmp/collector-config.yaml
+
+# run collector with the generated config
+cid=$(docker run -d --rm --name smoke-collector \
+    -v ./tmp/collector-config.yaml:/etc/otelcol-contrib/config.yaml \
+    otel/opentelemetry-collector-contrib:latest)
+sleep 1
```
TylerHelmuth, line 42: Can we test with our collector build instead so that we dont have to remove the components?

MikeGoldsmith: I was trying to do that but our supervised-collector docker image doesn't support local collector config files yet. I'm working on updating our supervisor image, then we update here later.

codeboten: the latest image has been pushed, should be good to go now

JamieDanielson (commented):
I think we should try to keep collector and refinery in separate test targets for now and see if we can get one reliably testing before adding the other. This is already hugely improved! If I comment out the collector parts in the smoke test script, and run `make smoke` I can already see that there is a problem with our EMAThroughputSampler. So even if we can run just the refinery pieces we can see which components need updating and then be able to add to CI to confidently know if there is a regression.

JamieDanielson:
> I can already see that there is a problem with our EMAThroughputSampler. So even if we can run just the refinery pieces we can see which components need updating and then be able to add to CI to confidently know if there is a regression.

I created a bug report for the EMAThroughputSampler ([internal ticket](https://github.com/honeycombio/pipeline-team/issues/261), probably could've gone here 🤷 )

MikeGoldsmith:
Moving back to draft until HPSF component issues are resolved.

JamieDanielson (approved):

PR #86, opened 2025-04-28
fix: Update EMA throughput component to produce valid configs

MikeGoldsmith:

- closes internal issue https://github.com/honeycombio/pipeline-team/issues/261

Updates the EMA throughput component to produce valid configs that Refinery can use to run.

- Update GoalThroughput to GoalThroughputPerSec
- Update AdjustmentInterval to be a duration with a valid defualt
- Update associated test data


pkg/data/components/EmaThroughput.yaml
```diff
@@ -55,8 +55,8 @@ properties:
       traffic patterns, but the more volatile the sampling rate will be.
     type: int
     validations:
-      - positive
-    default: 60
+      - duration
```
kentquirk, line 58: This isn't currently valid, because we don't yet have a duration editor in the UI, and type "int" will prevent entry of the duration suffix.

We can change the type to string, but I would prefer that we leave it as an int and append the "s" in the template.

Also, I'm in the middle of a redesign of all the refinery stuff, so please don't stress about this component too much. It's almost certainly changing rather radically.

MikeGoldsmith: Yeah, makes sense to use int and append the s as part writing the config. I'll update it. Thanks.

MikeGoldsmith: I've reverted back to using the positive validation. I got it to work with appending the `s` as a suffix, but had to remove the `encodeAsInt` operator otherwise it would fail to write the value correctly.

Example output:
```
Diff:
    --- Expected
    +++ Actual
    @@ -4,3 +4,3 @@
    EMAThroughputSampler:
    -            AdjustmentInterval: 60s
    +            AdjustmentInterval: "int\x1D60s"
```

kentquirk (approved):

PR #88, opened 2025-04-28
feat: Add smoke tests for collector configs using HPSF templates

MikeGoldsmith:

Follow-up to #84 where Refinery smoke tests were aded. This adds collector smoke tests.

- closes https://github.com/honeycombio/pipeline-team/issues/260

- Add `.smoke_collector` make file targets that generates collector configs and runs the collector in docker, checking it can start successfully
- Update EMA throughput template to provide a default FieldList which is a require property

Related PRs:
- https://github.com/honeycombio/hpsf/pull/86
- https://github.com/honeycombio/hpsf/pull/87


JamieDanielson (commented):
We chatted about narrowing the scope of this to only look at templates instead of all the components, as that should be a fully valid config. Later we can add more as we sort out what variations and situations we want to test. 👍

codeboten (approved):
would be good to add a check for yq version to ensure it works, it can be done in a follow up PR

Makefile
```diff
@@ -92,11 +92,59 @@ validate_all: examples/hpsf* pkg/data/templates/*
...
+	fi
+
+	@echo generating collector configs for component $(FILE)
+	mkdir -p tmp
+
+	# generate the configs from the provided file
+	go run ./cmd/hpsf -i ${FILE} -o tmp/collector-config.yaml cConfig || exit 1
+
+	# use yq to remove the usage processor and honeycomb extension from collector config
+	yq -i e \
```
JamieDanielson, line 111: Can you confirm my understanding? This removal of usage processor and honeycomb extension is still needed because we don't add these to the collector config until later in the process. Those are not available in this generated config and that's what we're testing, but they are expected in our supervised collector image and so we get failures if we keep those in. However, we are at least confirming that we can run our own image of the collector build with our generated configs.

JamieDanielson (approved):
Left a question for my own understanding, thanks for working on this!

Makefile
JamieDanielson: Re-reviewing this and testing with @codeboten , this is correct. The Honeycomb extension uses the opamp extension, so if that is not found, we get an error about a missing opamp extension.
`collector server run finished with error: failed to start extensions: extension "opamp" does not exist`
If we remove honeycomb and try to leave just the usage processor, we get an error because the usage processor is expecting to find the honeycomb extension. So both of these do still need to be removed.
`collector server run finished with error: cannot start pipelines: failed to start "usage" processor: extension "honeycomb" does not exist`

MikeGoldsmith: Yes and no 😄 

The usage processor requires the honeycomb extension, which requires the opamp extension. Configuring the the opamp extension is complicated because it requires an opamp server (eg beekeeper) to talk to. Removing these components means we don't need the additional setup and can just test the HPSF components without the config tooling chain.

PR #89, opened 2025-05-09
fix: update env var for refinery key

codeboten:
This updates the default variable for the refinery environment variable to the same value as the helm chart.


TylerHelmuth:
I think we should update the helm chart to match instead. There might be other places in HPSF we want to have a default exporting API key, and using a consistent variable name in HPSF makes sense

PR #90, opened 2025-05-16
feat: change the way rules files are generated

kentquirk:

- Rules files needed to be changed so that they're composable. The old way didn't work. This changes how they're generated and adds some new semantics so that we can build useful rules components. The problem is in the weirdly nested nature of refinery rules -- I wanted to avoid a structure that was tightly tied to refinery's current configuration structure but is still flexible enough to handle it. It's definitely got some rather refinery-specific code in here but it's not quite as brittle as an explicit structure.

- Add a rules_config config type (it uses dotted config but has some rules-specific fields and has extra semantics)
- Add a compact way of expressing rules conditions and a special way to handle them in the rules generator
- Add the ability to specify an index so we can create multiple rows
- Add a single rules-specific component
- Tweak the existing samplers to use the new rules_config
- Write some tests 
- Fix up some existing tests (for a couple of tests driven by control files, I added code to allow rewriting the control files within the test).
- Add some docs

### DIFF
--- a/pkg/config/rulestemplate.go
+++ b/pkg/config/rulestemplate.go
@@ -2,6 +2,7 @@ package config
 
 import (
 	"fmt"
+	"strconv"
 	"strings"
 
 	"github.com/honeycombio/hpsf/pkg/config/tmpl"
@@ -14,6 +15,7 @@ type rulesTemplate struct {
 }
 
 type rulesCondition struct {
+	index    int
 	fields   []string
 	op       string
 	value    string
@@ -23,27 +25,32 @@ type rulesCondition struct {
 func (r *rulesCondition) Render() (map[string]any, error) {
 	// render the condition into a dottedConfigTemplateKV
 	dc := make(map[string]any)
-	if len(r.fields) == 1 {
-		dc["Conditions.Field"] = r.fields[0]
-	} else {
-		dc["Conditions.Fields"] = r.fields
+	// we need to inject the index if it's not negative
+	c := "Conditions"
+	if r.index >= 0 {
+		c = fmt.Sprintf("Conditions[%d]", r.index)
 	}
-	dc["Conditions.Operator"] = r.op
+	if len(r.fields) == 1 {
+		dc[c+".Field"] = r.fields[0]
+	} else {
+		dc[c+".Fields"] = r.fields
+	}
+	dc[c+".Operator"] = r.op
 	if r.op == "exists" || r.op == "not_exists" {
 		// if the operator is exists or not_exists, we don't need a value or datatype
 		return dc, nil
 	}
-	dc["Conditions.Value"] = r.value
+	dc[c+".Value"] = r.value
 
 	switch r.datatype {
 	case "string", "s":
-		dc["Conditions.Datatype"] = "string"
+		dc[c+".Datatype"] = "string"
 	case "int", "i":
-		dc["Conditions.Datatype"] = "int"
+		dc[c+".Datatype"] = "int"
 	case "float", "f":
-		dc["Conditions.Datatype"] = "float"
+		dc[c+".Datatype"] = "float"
 	case "bool", "b":
-		dc["Conditions.Datatype"] = "bool"
+		dc[c+".Datatype"] = "bool"
 	case "":
 		// if the datatype is empty, don't set it
 	default:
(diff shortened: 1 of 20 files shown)


pkg/config/tmpl/dottedconfig_test.go
```diff
@@ -50,3 +50,54 @@ func TestMakeDottedConfig(t *testing.T) {
 		})
 	}
 }
+
+func clone(m map[string]any) map[string]any {
```
MikeGoldsmith, line 54: Could we use [maps.Clone](https://pkg.go.dev/maps@master#Clone) instead?

pkg/config/tmpl/rulesconfig.go
```diff
@@ -0,0 +1,73 @@
...
+func (rc *RulesConfig) RenderYAML() ([]byte, error) {
+	m := rc.RenderToMap(nil)
+	data, err := y.Marshal(m)
+	if err != nil {
+		return nil, err
+	}
+	return data, nil
+}
+
+func (rc *RulesConfig) Merge(other TemplateConfig) TemplateConfig {
```
MikeGoldsmith, line 53: Should we return a bool or error to indicate if the merge failed?

pkg/translator/testdata/refinery_rules/sampleerrors_defaults.yaml
```diff
@@ -0,0 +1,25 @@
...
+                  Name: Sample 500 statuses at 1
+                  SampleRate: "1"
+                - Conditions:
+                    - Datatype: int
+                      Fields:
+                        - http.status_code
+                        - http.response.status_code
+                      Operator: '>='
+                      Value: "400"
+                  Name: Sample 400 statuses at <no value>
```
MikeGoldsmith, line 22: I think the name has got a little messed up here - should be `Sample 400 statuses at 10`. Same happened for Sample Remainder below, but not for 500s.

MikeGoldsmith (approved):
This looks really good. I'd like to see a more complex refinery rules setup, maybe with multiple components what feed into the the same rules output but that can come later.

I've left some minor suggestions and queries but happy for this to move forward so we can test it and make tweaks later.

pkg/config/rulestemplate.go
```diff
@@ -0,0 +1,225 @@
...
+			case "f":
+				cond.fields = append(cond.fields, v.(string))
+			case "fs":
+				cond.fields = v.([]string)
+			case "o":
+				cond.op = v.(string)
+			case "v":
+				cond.value = v.(string)
+			case "d":
+				cond.datatype = v.(string)
```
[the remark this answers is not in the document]
kentquirk: It can only be a string or list of strings because it was generated above by splitting a string, and any failure is a component design error and would be caught immediately. I have updated the code to make it clearer.

pkg/config/rulestemplate.go
```diff
@@ -0,0 +1,225 @@
...
+			r.sampler, ok = mv.(string)
+			if !ok {
+				return r, fmt.Errorf("expected string for sampler, got %T", mv)
+			}
+		default:
+			return r, fmt.Errorf("unknown meta key %q", mk)
+		}
+	}
+
+	if r.env == "" {
```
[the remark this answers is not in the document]
kentquirk: Good catch.

pkg/config/rulestemplate.go
```diff
@@ -0,0 +1,225 @@
...
+		if err != nil {
+			return nil, err
+		}
+		rt.sampler = sampler
+	}
+
+	keyPrefix := "Samplers." + rt.env + "." + rt.sampler + "."
+	for _, kv := range rt.kvs {
+		// do the key
+		key, err := t.expandTemplateVariable(kv.key, userdata)
```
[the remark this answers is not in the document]
kentquirk: Yeah, turns out that's unnecessary, so I've deleted the other checks.

pkg/config/tmpl/dottedconfig.go
```diff
@@ -18,29 +42,98 @@ func (dc DottedConfig) renderInto(m map[string]any, key string, value any) {
...
+			// if the value is a map, we need to recursively call processIndices on it first
+			cv := processIndices(v)
+			if !pat.MatchString(k) {
+				// if the key doesn't match our regex, just add it to the map
+				out[k] = cv
+			} else {
+				// we need to process it -- split the key and index
+				matches := pat.FindStringSubmatch(k)
+				key := matches[1]
+				index, _ := strconv.Atoi(matches[2])
```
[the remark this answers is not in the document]
kentquirk: No need, since we know that matches[1] only contains digits.

pkg/config/tmpl/dottedconfig_test.go
kentquirk: No, because we need a deep clone, not a shallow one.

pkg/translator/testdata/refinery_rules/sampleerrors_defaults.yaml
kentquirk: Good catch.

TylerHelmuth (approved):

PR #92, opened 2025-05-27
fix: TraceConverter default values

TylerHelmuth:

- fixes https://github.com/honeycombio/pipeline-team/issues/383

We found the following issues with the traceconverter default install:
1. it uses the `{STRAWS_REFINERY_POD_IP}` env var that the collector does not have access to. 
2. Sets default otlp http port to 4318, but refinery listens for otlp http on port 80
3. Configures the otlphttp exporter to communicate using TLS, but Refinery is not using TLS.
4. The property is named Host, which implies a schema (http/https) does not need to be provided, but the default does not provide it.

- Updated env var to `${STRAWS_REFINERY_SERVICE}`. We'll update the helm chart to set this env var using the refinery service name it generates: https://github.com/honeycombio/helm-charts/pull/488
- update default port to 80
- update default TLS to be insecure. We changed the property name to be `Secure` instead of `Insecure` so a default value of `false` meant that the exporter would be rendered with `insecure: true`. We needed to inverted it since setting a boolean property to `false` via the HPSF config is not possible; the engine sees that as "not set".
- update rendered endpoint to set `http`/`https` based on `Secure` property.

### DIFF
--- a/pkg/data/components/TraceConverter.yaml
+++ b/pkg/data/components/TraceConverter.yaml
@@ -35,17 +35,18 @@ properties:
     type: string
     validations:
       - noblanks
-    default: ${STRAWS_REFINERY_POD_IP}
+    default: ${STRAWS_REFINERY_SERVICE}
     advanced: true
   - name: Port
     summary: The port on which Refinery is listening for HTTP traffic.
     description: |
       The port on which Refinery listens for incoming HTTP traffic.
-      The OTel default is 4318. Set to 0 to disable HTTP.
+      The default is 80. It is recommended not to change the default unless
+      you know what you're doing.
     type: int
     validations:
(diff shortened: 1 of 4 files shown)


pkg/data/components/TraceConverter.yaml
```diff
@@ -54,10 +55,11 @@ properties:
       configured. This properties supports sending a map of header keys and
       values.
     type: map
-  - name: Insecure
-    summary: Provide a way to disable TLS export.
+  - name: Secure
```
robbkidd, line 58: @kentquirk offered a good alternative name for the option: `UseTLS`. It's like the no-judgment name for this setting.

kentquirk (approved):

PR #94, opened 2025-05-28
feat: Simplify using default values in components

kentquirk:

- In order to handle defaults, we were relying on a magic incantation in components: `firstNonzero v1 v2 v3` with a particular pattern of values based on default values, user-specified values, and values coming from the environment. This PR replaces all of that with a single helper method on the TemplateComponent, `Data()`, which does all of that work in one reference.
- It also removes the now-obsolete `firstNonzero`, because if we need it again, it turns out it's pretty much just the Go template's `or` operator.

- Add Data method
- Replace firstNonzero references in a bunch of components
- Remove firstNonzero function
-


pkg/config/templateComponent.go
```diff
@@ -185,6 +185,37 @@ func (t *TemplateComponent) Props() map[string]TemplateProperty {
 	return props
 }
 
+func (t *TemplateComponent) Data() map[string]any {
```
TylerHelmuth, line 188: I'm tempted to call this `Values` to mimic helm, but thats just bc that templating engine is what I'm used to.

kentquirk: I like that better! i'm going to make that change.

PR #96, opened 2025-05-29
maint: Use numbers wherever possible in sampler

kentquirk:

- The sampler was generating a lot of strings where numbers should have been. This fixes that

- Use encodeAsInt for sample rates
- Make the condition encoder smarter
- Don't insist on strings in the templates
- Fix a test


codeboten (approved):
🎉

PR #97, opened 2025-05-29
feat: use batch config in exporters

codeboten:

- This removes the need for the separate batch processor and instead uses the batching configuration on the OTLP exporters directly. It cleans up the code used to inject the batch processor.

- update templates for otel grpc/http exporters and traceconverter
- update the tests

### DIFF
--- a/pkg/data/components/OTelGRPCExporter.yaml
+++ b/pkg/data/components/OTelGRPCExporter.yaml
@@ -59,6 +59,36 @@ properties:
     type: bool
     default: false
     advanced: true
+  - name: BatchTimeout
+    summary: How long to wait to before sending a batch, regardless of size.
+    description: |
+      Configure how long to wait before sending a batch. The batch will be sent after
+      this timeout.
+    type: duration
+    default: 60s
+    validations:
+      - duration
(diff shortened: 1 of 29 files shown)


pkg/data/components/OTelGRPCExporter.yaml
```diff
@@ -59,6 +59,36 @@ properties:
...
+  - name: QueueSize
+    summary: The size of a exporting queue.
+    description: |
+      The size of the exporting queue, measured by span/datapoint/log record count.
+      Items will be kept in the queue while the batch is being created.
+    type: int
+    default: 1_000_000
+    validations:
+      - nonempty
+    advanced: true
```
TylerHelmuth, line 91: These default made sense for the awss3exporter for data activation, but are likely too large for other things. Lets set the defaults to match the batch processor defaults:

- max size: 8192
- flush timeout: 200ms
- queue size: probably something much larger than 8192. maybe 100_000?

codeboten: updated defaults

TylerHelmuth (approved):

PR #98, opened 2025-05-30
fix: batch setting should be nested under batch section

TylerHelmuth:

- we added batching, but not under the correct section

- move batching under the `batch` section: https://github.com/open-telemetry/opentelemetry-collector/tree/main/exporter/exporterhelper#configuration


codeboten (approved):
one more test needs fixing

PR #100, opened 2025-06-01
feat: new rules components

kentquirk:

- We needed more examples of rules components
- There were some test problems with existing rules tests
- We were enforcing value's type before evaluation of templates; this was wrong. I removed it for now. A separate PR will add it back soon.

- Add some new rules components
   - KeepSlowTraces
   - KeepErrors
- Fix some test problems
   - Test file for SampleErrors was completely wrong 
- Remove type enforcement before template expansion

### DIFF
--- a/pkg/data/components/KeepErrors.yaml
+++ b/pkg/data/components/KeepErrors.yaml
@@ -33,9 +33,8 @@ properties:
   - name: FieldName
     summary: The name of the field to check for errors.
     description: |
-      The name of the field to check. If this field is not specified,
-      the default is 'error'. If this field exists, the trace will be
-      sampled at the specified rate.
+      The name of the field to check (default 'error'). If a field of this name
+      exists, the trace will be sampled at the specified rate.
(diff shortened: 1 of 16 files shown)


MikeGoldsmith (approved):

pkg/data/components/KeepErrors.yaml
```diff
@@ -0,0 +1,63 @@
...
+properties:
+  - name: Environment
+    summary: The environment in which to enable the sampler.
+    description: |
+      The environment in which to enable the sampler. If this field is
+      not specified, the sampler will be enabled in the __default__ environment,
+      which is what is used if the environment specified in the trace is not found.
+    type: string
+    default: "__default__"
+  - name: FieldName
```
MikeGoldsmith, line 33: Do you think it would be useful to have an array of field names in case there are multiple fields an error could be found? This might be useful for distributed traces where different processors / SDKs use different span attribute keys.

PR #106, opened 2025-06-06
test: Added end to end framework for config rendering

martinjt:

The HPSF rendering only checks for valid YAML, but not valid configs. The smoke tests can and will get long to run, this is a much faster to way to achieve the same goal.

Added a separate go module that encompasses the tests to ensure that dependencies don't get included.


Makefile
```diff
@@ -20,6 +20,7 @@ test_all: test_results
 	@echo "+++ testing - all the tests"
 	@echo
 	$(call GOTESTCMD,$@) -tags all --timeout 60s -v ./...
+	cd tests && $(call GOTESTCMD,$@) -tags all --timeout 60s -v ./...
```
TylerHelmuth, line 23: Can you break this out into its own make command and then use it here

tests/go.mod
TylerHelmuth, line 1: Was there an advantage to making `test` its own go module?

tests/providers/collector/components.go
```diff
@@ -0,0 +1,43 @@
...
+		},
+		Exporters: map[component.Type]exporter.Factory{
+			component.MustNewType("awss3"):    awss3exporter.NewFactory(),
+			component.MustNewType("debug"):    debugexporter.NewFactory(),
+			component.MustNewType("otlp"):     otlpexporter.NewFactory(),
+			component.MustNewType("otlphttp"): otlphttpexporter.NewFactory(),
+		},
+		Processors: map[component.Type]processor.Factory{
+			component.MustNewType("filter"):       filterprocessor.NewFactory(),
+			component.MustNewType("groupbyattrs"): groupbyattrsprocessor.NewFactory(),
```
TylerHelmuth, line 35: Do we have any hpsf components the use `groupbyattrs`?

tests/providers/collector/components.go
```diff
@@ -0,0 +1,43 @@
...
+	return otelcol.Factories{
+		Receivers: map[component.Type]receiver.Factory{
+			component.MustNewType("otlp"): otlpreceiver.NewFactory(),
+		},
+		Exporters: map[component.Type]exporter.Factory{
+			component.MustNewType("awss3"):    awss3exporter.NewFactory(),
+			component.MustNewType("debug"):    debugexporter.NewFactory(),
+			component.MustNewType("otlp"):     otlpexporter.NewFactory(),
+			component.MustNewType("otlphttp"): otlphttpexporter.NewFactory(),
+		},
```
TylerHelmuth, line 32: Should also include `nop`

tests/go.mod
martinjt: The idea was that it has a tonne of extra dependencies, and after some chat in the go channel, there wasn't confidence that it would bring them in.

Keeping it completely separate seemed like a better choice, especially considering that we shouldn't be using any of the internals as it's expected to replicate an outside, end-to-end scenario.

Makefile
martinjt: sure.

TylerHelmuth (approved):

PR #124, opened 2025-06-10
feat(components): Added JSON Schema file with vscode setting to enable it

martinjt:

Knowing what is valid for the names of the elements in a component can get annoying.

This adds a json schema based on the current setup, and also adds the vscode settings so that there is an extension available to do the validation and also setup to validate the files

### DIFF
--- a/component-schema.json
+++ b/component-schema.json
@@ -85,6 +85,35 @@
         "$ref": "#/$defs/TemplateProperty"
       }
     },
+    "validations": {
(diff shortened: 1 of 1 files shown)


MikeGoldsmith (approved):
Looks good. I updated the schema to comply with change since it was first introduced.

I did also try to add automatic schema validation with a makefile target, but couldn't get the tooling to work. Will try in another PR.

PR #125, opened 2025-06-10
feat: move dev component to alpha

MikeGoldsmith:

Updates the following components to `alpha` status which means they will be useable by more pipeline users.

- FilterBySeverityProcessor
- FilterProcessor
- KeepErrors
- LogDedupProcessor
- ParseAttributesAsJSON
- ParseLogBodyAsJSON

- Update component status to `alpha`
- Updates KeepError's port names for consistency (Events)

### DIFF
--- a/pkg/data/components/FilterProcessor.yaml
+++ b/pkg/data/components/FilterProcessor.yaml
@@ -2,7 +2,7 @@ kind: FilterProcessor
 name: Filter Processor
 style: processor
 type: base
-status: development
+status: alpha
 version: v0.0.0
 summary: A processor that can be used to filter telemetry.
 description: |
--- a/pkg/data/components/FilterProcessor.yaml
+++ b/pkg/data/components/FilterProcessor.yaml
@@ -2,7 +2,7 @@ kind: FilterProcessor
 name: Filter Processor
 style: processor
 type: base
-status: alpha
+status: development
 version: v0.0.0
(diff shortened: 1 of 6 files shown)


pkg/data/components/FilterProcessor.yaml
```diff
@@ -2,7 +2,7 @@ kind: FilterProcessor
 name: Filter Processor
 style: processor
 type: base
-status: development
+status: alpha
```
martinjt, line 5: This shouldn't be alpha, it's broken and not required

pkg/data/components/KeepErrors.yaml
```diff
@@ -2,7 +2,7 @@ kind: KeepErrors
 name: Keep Errors
 style: sampler
 type: base
-status: development
+status: alpha
```
martinjt, line 5: This isn't the right component, the SampleByHTTPErrors is already done

pkg/data/components/LogDeduplicationProcessor.yaml
```diff
@@ -2,7 +2,7 @@ kind: LogDeduplicationProcessor
 name: Log Deduplication Processor
 style: processor
 type: base
-status: development
+status: alpha
```
martinjt, line 5: This wasn't on the list, but maybe @verythorough wants it?

martinjt (approved):

PR #132, opened 2025-06-12
fix: resolve and render pipelines individually. 

kentquirk:

Reworks HPSF to generate pipelines one signal at a time, and then merge them later. It passes all tests in both the main and tests modules except the one called `multiple_oltp_exporters`. For that test, it generates separate pipelines rather than combining them into one exporter. I think that's OK for now, so I've skipped it and we can fix it in a different PR.

- generate individual pipelines
- resolve them individually by signal type
- because of complications related to pipeline indexing (`traces/1`) we now hash the individual pipeline data and produce a deterministic hex value for a given pipeline (`traces/2ac5`)
- fix a gazillion tests
- clean up a few things that were messy
- skip the multiple_otlp_exporters test

### DIFF
--- a/pkg/config/templateComponent.go
+++ b/pkg/config/templateComponent.go
@@ -235,7 +235,7 @@ func (t *TemplateComponent) ConnectsUsingAppropriateType(connType hpsf.Connectio
 // // ensure that TemplateComponent implements Component
 var _ Component = (*TemplateComponent)(nil)
 
-func (t *TemplateComponent) GenerateConfig(cfgType Type, pipeline hpsf.PipelineWithSignalType, userdata map[string]any) (tmpl.TemplateConfig, error) {
+func (t *TemplateComponent) GenerateConfig(cfgType Type, pipeline hpsf.PipelineWithConnectionType, userdata map[string]any) (tmpl.TemplateConfig, error) {
 	// we have to find a template with the kind of the config; if it
 	// doesn't exist, we return an error
 
@@ -271,7 +271,7 @@ func (t *TemplateComponent) GenerateConfig(cfgType Type, pipeline hpsf.PipelineW
 			case "rules":
 				// a rules template expects the metadata to include environment
 				// information.
-				if pipeline.SignalType != hpsf.CTYPE_HONEY {
+				if pipeline.ConnType != hpsf.CTYPE_HONEY {
 					continue // rules templates are only for Honeycomb events
 				}
 				rt, err := buildRulesTemplate(template)
@@ -410,7 +410,7 @@ func (t *TemplateComponent) applyTemplate(tmplVal any, userdata map[string]any) 
 
 // this is where we do the actual work of generating the config; this thing knows about
 // the structure of the collector config and how to fill it in
-func (t *TemplateComponent) generateCollectorConfig(ct collectorTemplate, pipeline hpsf.PipelineWithSignalType, userdata map[string]any) (*tmpl.CollectorConfig, error) {
+func (t *TemplateComponent) generateCollectorConfig(ct collectorTemplate, pipeline hpsf.PipelineWithConnectionType, userdata map[string]any) (*tmpl.CollectorConfig, error) {
 	// we have to fill in the template with the default values
 	// and the values from the properties
 	t.collName = ct.collectorComponentName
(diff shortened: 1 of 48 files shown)


pkg/config/templateComponent.go
```diff
@@ -270,14 +263,17 @@ func (t *TemplateComponent) GenerateConfig(cfgType Type, userdata map[string]any
...
-				tmpl, err := t.generateCollectorConfig(ct, userdata)
+				tmpl, err := t.generateCollectorConfig(ct, pipeline, userdata)
 				if err != nil {
 					return nil, err
 				}
 				generatedTemplates = append(generatedTemplates, tmpl)
 			case "rules":
 				// a rules template expects the metadata to include environment
 				// information.
+				if pipeline.SignalType != hpsf.CTYPE_HONEY {
```
VinozzZ, line 274: Ohh, is this what fixed the problem with duplicated rules generation?

pkg/hpsf/hpsfTypes.go
```diff
@@ -526,6 +559,117 @@ func (h *HPSF) GetStartComponents() []*Component {
...
+	}
+	buf.WriteString(string(p.SignalType))
+	hash := metro.Hash64(buf.Bytes(), 0x234da488) // use a fixed seed for reproducibility
+	shash := strconv.FormatUint(hash, 16)
+	return shash[len(shash)-4:] // return the last 4 characters of the hash
+}
+
+// FindAllPipelines generates all paths from the start components to the end
+// components where end components are those that are not sources of any
+// connections. It returns a slice of slices of components, where each inner slice
```
VinozzZ, line 626:
```suggestion
FindAllPipelines generates all paths for a given signalType from the start components to the end
// components where end components are those that are not sources of any
// connections.
```

pkg/translator/translator.go
```diff
@@ -248,25 +254,66 @@ func (t *Translator) GenerateConfig(h *hpsf.HPSF, ct config.Type, userdata map[s
...
 	}
 
-	// Start with a base component so we always have a valid config
-	dummy := hpsf.Component{Name: "dummy", Kind: "dummy"}
-	base := config.GenericBaseComponent{Component: dummy}
-	composite, err := base.GenerateConfig(ct, userdata)
-	if err != nil {
-		return nil, err
+	// We need to generate our collection of unique pipelines. A pipeline in
+	// this context is the shortest path from a source component to a
```
VinozzZ, line 258: Maybe this comment can also be added for the type definition of `pipelineWithSignalType`

pkg/hpsf/hpsfTypes.go
```diff
@@ -17,16 +19,46 @@ type ConnectionType string
...
+	CTYPE_TRACES  ConnectionType = "OTelTraces"
+	CTYPE_EVENTS  ConnectionType = "OTelEvents"
+	CTYPE_HONEY   ConnectionType = "HoneycombEvents"
 	CTYPE_NUMBER  ConnectionType = "number"
 	CTYPE_STRING  ConnectionType = "string"
 	CTYPE_BOOL    ConnectionType = "bool"
 )
 
+// These are the possible connection types that can be used in the HPSF pipelines.
+var PipelineSignalTypes = []ConnectionType{
```
MikeGoldsmith, line 33: We've deviated from connection to signal type in these new variables and I found it to be confusing. Can we try to consolidate on a single term? I'm okay with either Connection or Signal.

pkg/hpsf/hpsfTypes.go
```diff
@@ -526,6 +559,117 @@ func (h *HPSF) GetStartComponents() []*Component {
...
+	// return the ID of the pipeline, which is a hash of the names of components in its pipeline
+	// and the signal type, truncated to 4 characters.
+	buf := bytes.Buffer{}
+	for _, comp := range p.Pipeline {
+		buf.WriteString(comp.GetSafeName())
+	}
+	buf.WriteString(string(p.SignalType))
+	hash := metro.Hash64(buf.Bytes(), 0x234da488) // use a fixed seed for reproducibility
+	shash := strconv.FormatUint(hash, 16)
+	return shash[len(shash)-4:] // return the last 4 characters of the hash
```
MikeGoldsmith, line 621: What is the probability of two pipelines generating a hash where the last 4 characters match and would clash?

pkg/config/templateComponent.go
kentquirk: Yes! We were generating rules for every pipeline.

martinjt:
The biggest issue I see with this change is the increased memory and CPU usage through duplicating pipelines. At scale, that could be quite large depending on the number and type of processors they add.

If we are going to follow this up with a fast follow on consolidating pipelines, then that's probably fine. If we're not going to consolidate the pipelines before GA, ensure there is a note somewhere about this as it will be something customers may recognise and question. For Beta, we should make the SA/CA teams aware of the limitation.

On the test that's been skipped, we should fix that test so it represents what's happening.

There's also the wider issue of the usage processor duplicating data. This wouldn't be fixed by consolidating pipelines, but I have a change that should solve that part.

I have a branch that I'll PR after this is merged which adds all the relevant multiple pipeline scenarios.

pkg/hpsf/hpsfTypes.go
kentquirk: Good point. Connection is baked into HPSF so I'll keep that one.

pkg/hpsf/hpsfTypes.go
kentquirk: There are 65536 possible 4-letter hashes. There is a one in 10,000 chance of collision when there are 5 pipelines in the mix. Even if there are 500 pipelines it's still well under 1%. 

HPSF works pretty hard to make its generated configs both legible and idiomatic by humans. Long hex strings are not something normal humans are comfortable with. 

I changed it to 6-character, with a dash, which drops the collision probability to 1 in a million for 4 pipelines, 1 in 10,000 for 500 pipelines.

PR #133, opened 2025-06-12
test: Update Tests to support named pipelines

martinjt:

Test tests assumed the first pipeline will not have a suffix

Adds a helper so that you can get the first pipeline of a type


martinjt:
some errorneous commits on this, checking

kentquirk (approved):
I'll be honest, I haven't reviewed this in detail, but I rebased my branch on it and it passes, so I'm happy with it.

PR #134, opened 2025-06-13
test: Added test for the various multiple pipeline scenarios

martinjt:

No validation that HPSF definitions that should generate distinct pipelines is actually doing that.

This is only tests to ensure that changes don't break that functionality. These should be resilient to pipeline consolidation when that happens.


MikeGoldsmith (approved):
Tests look fine. The commented expected pipeline formatting need updating 😄

PR #136, opened 2025-06-13
fix(validation): Fixed the validation for properties that aren't supplied

martinjt:

Validation logic for non-supplied parameters wasn't working

- Inverted the loops for checking the properties so it uses the TemplateComponent as the source, and not the user supplied information.
- Added tests

### DIFF
--- a/pkg/config/templateProperty.go
+++ b/pkg/config/templateProperty.go
@@ -142,6 +142,10 @@ func positive(val any) bool {
 // ensures that no strings are blank, even in string slices
 // use nonempty to ensure that a slace has at least one value
 func noBlankStrings(val any) bool {
+	// nil happens when there is no default value and no value is supplied
+	if val == nil {
+		return false
+	}
 	switch v := val.(type) {
 	case string:
 		return len(v) > 0
--- a/pkg/translator/translator.go
+++ b/pkg/translator/translator.go
@@ -138,11 +138,14 @@ func (t *Translator) ValidateConfig(h *hpsf.HPSF) error {
 		}
 
 		// Get the template properties from the template component.
-		tprops := tmpl.Props()
+		templateProperties := tmpl.Props()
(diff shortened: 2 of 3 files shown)


martinjt:
I've had to make a few changes here, but one notable change needs thinking about.

HoneycombExporter required a `url` type, but the default was `https://api.honeycomb.io:443` (i.e. with the port), therefore even with the default being validated, this still failed. I've updated that, however, the same property was used in the `Network` segment of the refinery config, I'm not sure that is write, and we maybe need to update the refinery config rendering to include the port.

This also raises a point that for this exporter, the default should actually come from an Environment Variable, or it won't work on the EU. If you do add an Environment Variable, it won't pass validation. That means we likely need a new property type or overriding validation that means "Accepts Environment Variable". I was thinking that maybe, this is a special Validation type that's checked on it's own, outside of the loop, and if it detects and environment variable than it passes validation immediately (or maybe we just do that for all fields?).

I'll await feedback on the above before I merge.

pkg/data/components/HoneycombExporter.yaml
```diff
@@ -54,7 +54,6 @@ properties:
     type: string
     validations:
       - noblanks
-      - url
```
MikeGoldsmith, line 57: We could update the url validator to optionally allow a port.

MikeGoldsmith (approved):
I think this look good. Thanks @martinjt.

I've left a suggestion for updating the url validator. I think that could be done in another PR though so not to block this PR.

pkg/data/components/HoneycombExporter.yaml
MikeGoldsmith: Now https://github.com/honeycombio/hpsf/commit/af0c25b487cb45afd5483fee60d4a3c448cd09f1 is merged. We should be okay to keep this validator.

MikeGoldsmith (approved):
🎉

PR #149, opened 2025-06-23
feat: Add RemoveAttribute processor

MikeGoldsmith:

Adds a new processor that can remove attributes from traces, metrics and log records.

- Closes https://github.com/honeycombio/pipeline-team/issues/438

- Adds new `encodeAsArrayWithFormat` config helper
- Add new `RemoveAttributesProcessor`
- Add unit tests to verify behaviour

### DIFF
--- a/pkg/data/components/RemoveAttributeProcessor.yaml
+++ b/pkg/data/components/RemoveAttributeProcessor.yaml
@@ -48,12 +48,15 @@ properties:
     validations:
       - noblanks
       - nonempty
-  - name: RemoveResourceAttributes
-    summary: A flag that indicates whether to remove resource attributes.
+  - name: AttributeLocation
+    summary: A flag that indicates whether to remove resource or record attributes.
     description: |
-      A flag that indicates whether to remove resource attributes from the telemetry data.
-    type: boolean
-    default: false
+      A flag that indicates whether to remove resource or record attributes from the telemetry data.
+    type: string
+    sub_type: oneof(record, resource)
+    validations:
+      - oneof(record, resource)
+    default: record
 templates:
   - kind: collector_config
     name: otel_remove_attributes_processor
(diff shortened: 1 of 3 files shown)


pkg/translator/testdata/collector_config/removeattributeprocessor_defaults.yaml
```diff
@@ -0,0 +1,31 @@
+receivers:
+    otlp/otlp_in:
+        protocols:
+            grpc:
+                endpoint: ${STRAWS_COLLECTOR_POD_IP}:4317
+            http:
+                endpoint: ${STRAWS_COLLECTOR_POD_IP}:4318
+processors:
+    transform/remove_attributes_1:
+        error_mode: ignore
```
TylerHelmuth, line 10: This is an invalid transformprocessor config, it will cause the collector not to start.

TylerHelmuth (commented):
I believe we need to update validation so that it is not possible to add the processor with 0 specified attributes.

pkg/data/components/RemoveAttributeProcessor.yaml
```diff
@@ -0,0 +1,75 @@
...
+    meta:
+      componentSection: processors
+      signalTypes: [traces, metrics, logs]
+      collectorComponentName: transform
+    data:
+      - key: "{{ .ComponentName }}.error_mode"
+        value: "ignore"
+      # Trace statements
+      - key: "{{ .ComponentName }}.trace_statements"
+        value: "{{ .Values.Attributes | encodeAsArrayWithFormat \"delete_key(resource.attributes, \\\"%s\\\")\" }}"
```
TylerHelmuth, line 63: I do wonder if we need 1 more property that lets you toggle on/off Resource attributes. Or maybe we need a separate RemoveResourceAttributes component.

I am thinking of the use case where maybe an attribute is duplicate on both the resource and item and the user only wants to remove from 1. I fear always removing from the resource and the item will result in accidental removals.

pkg/data/components/RemoveAttributeProcessor.yaml
```diff
@@ -0,0 +1,75 @@
+kind: RemoveAttributeProcessor
+name: Remove Attribute Processor
```
TylerHelmuth, line 2: We dont need to include `processor` in the name displayed to the customer
```suggestion
name: Remove Attributes
```

pkg/data/components/RemoveAttributeProcessor.yaml
```diff
@@ -0,0 +1,75 @@
+kind: RemoveAttributeProcessor
```
TylerHelmuth, line 1:
```suggestion
kind: RemoveAttributesProcessor
```

pkg/data/components/RemoveAttributeProcessor.yaml
```diff
@@ -0,0 +1,85 @@
...
+    type: stringarray
+    validations:
+      - noblanks
+      - nonempty
+  - name: RemoveResourceAttributes
+    summary: A flag that indicates whether to remove resource attributes.
+    description: |
+      A flag that indicates whether to remove resource attributes from the telemetry data.
+    type: boolean
+    default: false
```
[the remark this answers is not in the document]
MikeGoldsmith: We could also use a oneof(record, resource, both) to allow all three scenarios?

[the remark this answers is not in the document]
TylerHelmuth: @verythorough what do you think?

PR #150, opened 2025-06-23
feat: Add RenameAttributes processor

MikeGoldsmith:

Adds a new processor that can rename attributes on span, metric and log records. It can optionally drop the old attribute names.

- Closes https://github.com/honeycombio/pipeline-team/issues/450

- Update the new `encodeAsArrayWithFormat` helper to support maps
- Add new RenameAttributes processor
- Add unit tests to check expected output


pkg/data/components/RenameAttributeProcessor.yaml
```diff
@@ -0,0 +1,76 @@
...
+    meta:
+      componentSection: processors
+      signalTypes: [traces, metrics, logs]
+      collectorComponentName: transform
+    data:
+      - key: "{{ .ComponentName }}.error_mode"
+        value: "ignore"
+      # Trace statements
+      - key: "{{ .ComponentName }}.trace_statements"
+        value: "{{ .Values.Attributes | encodeAsArrayWithFormat \"set(resource.attributes[\\\"%s\\\"], resource.attributes[\\\"%s\\\"])\" }}"
```
martinjt, line 64: You'll need where clauses on each of these to ensure the original attribute exists.

TylerHelmuth: Also on the delete, we'll only want to delete the old attribute if the new attribute exists.

MikeGoldsmith:
Closing for now. We have the generic TransformProcessor component that can be used until we decide we want to add more specialised components.

PR #159, opened 2025-07-09
feat: rewrite the way refinery samplers work

kentquirk:
# THIS IS A BREAKING CHANGE TO EXISTING HPSF FILES

I'm really sorry this couldn't be smaller, but it couldn't.

Sampler configurations created with previous versions of HTP weren't composable -- you couldn't do interesting sampler stuff. So we did some research, and came up with a new paradigm. This PR implements it.

Big picture changes:
* Samplers are connected with a new type of connection -- SampleData. It's currently gray because we haven't done anything better in the UI yet.
* StartSampling now begins a sampling chain. It has 4 outputs, and they are evaluated in order.
* Each chain consists of 0 or more conditions, each of which is ANDed together. If a trace being evaluated matches a condition, it moves on to the next.
* Each chain ends with a sampler. The sampler only decides on the probability of keeping that trace. We currently have keep, drop, deterministic and a couple of ema samplers. We can create more.
* If a condition fails, the trace moves on to the next output in StartSampling. We want to evaluate how that component should work (how many outputs should it have) but that's outside this discussion.
* Samplers (except Drop) then have an output that goes to the exporter.
* Lots of tests got changed/fixed. A couple of obsolete tests were deleted.
* Validations are not yet complete. There are things that should be checked that are not yet. We'll do that separately.

This system implements the ability for StartSampling to specify the environment name, to support multiple sets of rules the way Refinery does, but that's blocked by other work in HTP for now, so we don't use it yet.

## Still to do in future PRs:

- Consider whether some of these components or ports should have different names
- Enhance validations to include the new sampler rules
- build the rest of the sampler components
- build a bunch more condition components, including a generic one
- fix the readme documentation
- Add more tests for more sampler combinations
- Add a couple of scenario tests
- Generate the Name field in Refinery rules in a better way; today it's just the name of the sampler
- Dynamic handle creation in StartSampling, or multiple StartSampler components with different numbers of handles.
- Add env property to StartSampling (also requires other work to use it properly)


.cursorrules
TylerHelmuth, line 1: Should this be checked in?

examples/hpsf.yaml
TylerHelmuth, line 1: Do we have replacement examples? Or are these not useful anymore?

pkg/translator/translator.go
```diff
@@ -225,85 +227,90 @@ func (t *Translator) validateConnectionPorts(h *hpsf.HPSF, templateComps map[str
...
+	// 				samplerConnections++
+	// 			}
+	// 		}
+	// 	}
+	// 	if samplerConnections != 1 {
+	// 		err := hpsf.NewError("StartSampling component must have exactly one connection to a sampler").
+	// 			WithComponent(startSamplingComp)
+	// 		result.Add(err)
+	// 	}
+	// }
```
TylerHelmuth, line 313: Do we still need this code?

pkg/config/tmpl/rulesconfig.go
```diff
@@ -1,73 +1,212 @@
...
+func RCTFromStyle(style string) RulesComponentType {
+	switch style {
+	case "startsampling":
+		return StartSampling
+	case "condition":
+		return Condition
+	case "sampler":
+		return Sampler
+	case "dropper":
+		return Dropper
```
TylerHelmuth, line 29: casting the const RulesComponentTypes to a string here would be nice to ensure this mapping is not broken by a rename.

pkg/translator/translator.go
kentquirk: I left this in deliberately because this PR doesn't include fixing all the validation, but this code is still useful as reference.

pkg/config/tmpl/refineryrules.go
```diff
@@ -0,0 +1,260 @@
...
+	}
+}
+
+// SetMemberValue sets a value on a member of a struct, including nested structs, slices, and maps, dereferencing pointers as needed.
+// It also handles Duration type conversion.
+// This code uses reflection. The go proverb says "Clear is better than clever. Reflection is never clear."
+// This is an example of that, but the alternative is a whole bunch of type-specific code to do this for every possible type.
+// It was written with a lot of help from Claude, and passes its tests.
+// If this function returns an error, it is almost certainly due to a component design issue.
+func SetMemberValue(key string, member any, value any) error {
```
TylerHelmuth, line 185: Can this be unexported?

pkg/hpsf/hpsfTypes.go
```diff
@@ -620,15 +622,27 @@ func (h *HPSF) isSourceComponent(c *Component, connType ConnectionType) bool {
...
-// connected by a specific connection type. We generate all possible pipelines
-// as a slice of this type.
-type PipelineWithConnectionType struct {
-	ConnType ConnectionType
-	Pipeline []*Component
+// PipelineWithConnections is designed to hold a linear set of components
+// connected by a specific connection type, along with the specific
+// sets of connections used. We generate all possible pipelines
+// as a slice of these.
+type PipelineWithConnections struct {
```
TylerHelmuth, line 629: What does "pipeline" mean here? Is it the Pipeline Installation, the collector pipeline, or some other thing?

pkg/config/templateComponent.go
```diff
@@ -282,15 +299,30 @@ func (t *TemplateComponent) GenerateConfig(cfgType Type, pipeline hpsf.PipelineW
...
 				}
 				rt, err := buildRulesTemplate(template)
 				if err != nil {
 					return nil, fmt.Errorf("error %w building rules template for %s",
 						err, t.Kind)
 				}
-				tmpl, err := t.generateRulesConfig(rt, userdata)
+
+				// We need to know which pipeline we're generating rules for, so we look up the
+				// connection leading to this component.
```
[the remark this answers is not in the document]
kentquirk: After thinking about it, "path" is a better term -- it's a single path through the graph for a single connection type. I've renamed it and changed the comments.

pkg/hpsf/hpsfTypes.go
kentquirk: See above; I've renamed to "path".

PR #170, opened 2025-07-15
feat: Use the sampler name as the rule name when generating refinery configs

kentquirk:

- When generating HPSF rules for Refinery, it wasn't generating the Name field for the rules. This uses the sampler name as the rule name so that the rules can be documented. Since the names must be unique, this is safe and helpful.

- Pass the component name into the metadata
- When conditions are right, use the component name to extend the data for the rules
- Fix all the tests to look for rule names
- The keep and drop samplers had a RuleName field that wasn't being used or tested for. Remove it.

### DIFF
--- a/pkg/config/tmpl/rulesconfig.go
+++ b/pkg/config/tmpl/rulesconfig.go
@@ -3,6 +3,7 @@ package tmpl
 import (
 	"fmt"
 	"strconv"
+	"strings"
 
 	y "gopkg.in/yaml.v3"
 )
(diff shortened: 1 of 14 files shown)


pkg/config/tmpl/rulesconfig.go
```diff
@@ -169,6 +171,14 @@ func (rc *RulesConfig) Merge(other TemplateConfig) error {
 				return err
 			}
 		}
+		// Add the Name field if we're creating a rule (keyPrefix starts with "RulesBasedSampler.Rules." but doesn't contain "Conditions")
+		if strings.HasPrefix(keyPrefix, "RulesBasedSampler.Rules.") && !strings.Contains(keyPrefix, "Conditions") {
```
codeboten, line 175: not a blocker, but it would be good to have this in a separate function to reuse it in both places where this logic is needed

codeboten (approved):

pkg/config/tmpl/rulesconfig.go
kentquirk: I made it better. It doesn't remove all the duplication but when I tried to, it looked worse.

PR #182, opened 2025-07-21
fix: Better sampler validation (conforms to new sampler design).

kentquirk:

- The sampler component redesign needed to turn off some validations. This fixes them.

- Update sampler connection validations
- Fix some tests

### DIFF
--- a/pkg/translator/translator.go
+++ b/pkg/translator/translator.go
@@ -172,7 +172,7 @@ func (t *Translator) validateSamplerConnections(h *hpsf.HPSF, templateComps map[
 			continue
 		}
 
-		if tmpl.Style == "sampler" || tmpl.Style == "condition" {
+		if tmpl.Style == "sampler" || tmpl.Style == "dropper" || tmpl.Style == "condition" {
 			// check the connections for the component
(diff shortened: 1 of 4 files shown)


pkg/translator/translator.go
```diff
@@ -226,90 +231,120 @@ func (t *Translator) validateConnectionPorts(h *hpsf.HPSF, templateComps map[str
...
+	dfs(startComp)
+	return components
+}
+
+// The rules for sampling in HPSF are as follows:
+// - If there are any sampling components, there must be at least one component with style "startsampling".
+// - Each path connected to a "startsampling" component's output must lead to exactly one "sampler" or "dropper".
+// - There may be multiple "condition" components between startsampling and the sampler or dropper.
+// - Droppers can terminate a path (since they do not have an output port).
+// - The output of samplers must be connected to an "exporter" component.
```
TylerHelmuth, line 266: Is it also possible to add a rule for:

```
If a startsampling component has multiple outputs, all but 1 MUST have a condition
```

TylerHelmuth (approved):

PR #187, opened 2025-07-24
chore: clean up following linter suggestions

codeboten:

- Ran a linter locally against the repo and am following some of the suggestions to clean up the code a bit. No functional changes in this PR, though it did find a potential bug in main.go where a file handle could be left dangling, and simplified some of the indented cases that were not necessary.

- updated comments
- added a call to close a file handle when loading embedded components could return early
- removed a few unnecessary indentation blocks
- updated assert.Error to use require.Error
- use strconv where possible instead of Sprintf
-

### DIFF
--- a/cmd/hpsf/main.go
+++ b/cmd/hpsf/main.go
@@ -87,10 +87,11 @@ func main() {
 
 	// Create the output file
 	var outf io.Writer
+	var f *os.File
 	if cmdopts.Output == "-" {
 		outf = os.Stdout
 	} else {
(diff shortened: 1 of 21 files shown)


cmd/hpsf/main.go
TylerHelmuth, line 1: which linter did you use? can you throw a `make lint` target into the Makefile

TylerHelmuth: Dont add it to CI yet

TylerHelmuth (approved):

PR #195, opened 2025-07-30
feat: add library_version field, add FromYAML func

codeboten:

- Adding `library_version` field to store the version of the library that was used when generating HPSF. This version will be used later on when using different versions of the library to translate HPSF into various configurations.
- Introduce a `FromYAML` func to use downstream

- add LibraryVersion struct field 
- add FromYAML func and clean up various areas where decoding was done

### DIFF
--- a/cmd/hpsf/main.go
+++ b/cmd/hpsf/main.go
@@ -115,12 +115,12 @@ func main() {
 
 	switch cmds[0] {
 	case "format":
-		hpsf, err := unmarshalHPSF(inputRdr)
+		h, err := hpsf.FromYAML(inputRdr)
 		if err != nil {
 			log.Fatalf("error unmarshaling input file: %v", err)
 		}
 		// write it to the output file as yaml
-		data, err := y.Marshal(hpsf)
+		data, err := y.Marshal(&h)
 		if err != nil {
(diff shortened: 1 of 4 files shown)


kentquirk (commented):
I'm not sure I understand yet how LibraryVersion is expected to be used. Are you expecting the UI to set this value? And the HPSF library to check it? Can you please show a scenario of how this will play out in practice?

codeboten:
Right, whatever is writing the HPSF is expected to record the library version. This version can then be used at a later time to determine if the version of the HPSF library that is used to read the HPSF and translate it into configuration is compatible with the version that was used to write the config.

The scenario this is anticipating is where a breaking change in the HPSF library would require some kind of migration to the configuration data. Or in the case where no migration path is supported, an older version of the library could be used by whoever is reading the HPSF.

A user saves a config file today. With no version information about the library being stored, any version of the HPSF library would be trying to interpret that configuration as data it should understand but may fail to do so. 

With a version, the library could offer more information about why it failed to read the data, or if there's a path forward, it could potentially offer that as well. The caller of the library could also use this LibraryVersion to choose which library to use for the translation action.

kentquirk (approved):
Thanks for the explanation.

PR #197, opened 2025-07-30
feat: add type-specific comparison and scope to conditions

kentquirk:

- Adds condition components for comparison of different value types. 
- Adds the ability for a condition to set the `Scope` on a rule. If `Scope: span` is specified anywhere, it always wins.
- Adds a component to force `Scope: span` on conditions (only useful if you have more than one condition in a rule)
- For negative conditions (!=) in these comparisons, it forces `Scope: span` automatically, as there is no valid use case for `Scope: trace` with a negative condition.
- Adds/updates appropriate test cases, including a big and fairly exhaustive test for the scope rules.
- Adds these components to the everything_alpha test under a new Rule.


pkg/data/components/CompareDecimalField.yaml
```diff
@@ -0,0 +1,64 @@
+kind: CompareDecimalField
```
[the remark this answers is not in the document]
kentquirk: I think those guidelines need revisiting; that's a very long name.

pkg/translator/testdata/refinery_rules/forcespanscope_all.yaml
TylerHelmuth, line 1: It doesnt look like these scope changes rendered correctly

kentquirk: Did that.

pkg/config/tmpl/rulesconfig.go
```diff
@@ -275,6 +308,24 @@ func (rc *RulesConfig) Merge(other TemplateConfig) error {
...
+					if existingScope := sampler.RulesBasedSampler.Rules[ruleIndex].Scope; existingScope != "" {
+						currentScope = existingScope
+					}
+				}
+
+				mergedScope := mergeScopeValues(currentScope, scope)
+				if err := setMemberValue(fmt.Sprintf("RulesBasedSampler.Rules.%d.Scope", ruleIndex), sampler, mergedScope); err != nil {
+					return err
+				}
+			}
```
TylerHelmuth, line 327: Is this the same code as above? If so a shared function would be nice.

kentquirk: Similar, but not identical. I played with it a bit, but there are enough differences that it didn't feel worth it to try to combine them (the function would have several arguments).

TylerHelmuth (approved):

PR #198, opened 2025-07-31
feat: Regexp Component

kentquirk:

- Implements MatchRegularExpression as a Refinery condition (using the 'matches' operator).

- Add MatchRegularExpression
- Add a validation called regex that checks the validity of a regex
- Add tests

Dependent on #197 and should be merged after it.


pkg/data/components/MatchRegularExpression.yaml
```diff
@@ -0,0 +1,57 @@
...
+      - noblanks
+  - name: RegularExpression
+    summary: The regular expression to match against the field value.
+    description: |
+      The regular expression pattern that the field value should match. The regular expression must be a valid Go regular expression.
+      Example: `^[A-Z][A-Za-z0-9]+$` would match a capital letter followed by one or more letters or numbers.
+    type: string
+    validations:
+      - noblanks
+      - regex
```
TylerHelmuth, line 42: Based on the sync discussion this week should we provide a default value?

kentquirk: My takeaway was if we had a reasonable default we should -- but I couldn't think of a reasonable one.

PR #200, opened 2025-08-01
feat: Add FieldExistsCondition

kentquirk:

- This is a generic FieldExists component that supports testing whether a field exists or not.

- Add the component, including the logic to force scope: span when we use the negative condition
- Adjust All The Tests

Based on #198


codeboten (approved):

TylerHelmuth (commented):
Can you also add it to the e2e test

PR #202, opened 2025-08-05
feat: add anthropic receiver

austinlparker:
This adds a new component for the anthropic receiver.


TylerHelmuth:
As a general note about the receiver, I recommend following contrib's scrape pattern wherever possible. If we can add a metadata.yaml and use it to define the metrics and then use mdatagen to generate the metricsbuilder from the metadata.yaml we'll gain a lot of confidence in the component.

austinlparker:
Do you have an example of this? Given that the API shape is a bit unique, I'm not sure how much we can re-use existing stuff.

PR #214, opened 2025-08-28
feat: Add Apply OTel Semconv processor

MikeGoldsmith:

Adds a new processor component that can try to apply common k8s semconv values. Currently only configured to work for logs, could be expanded to include other signals later.

- closes https://github.com/honeycombio/pipeline-team/issues/602

- Add new processor
- Add unit tests
- Add scenario tests


pkg/data/components/ApplyK8sSemconvForLogs.yaml
```diff
@@ -0,0 +1,46 @@
...
+    format: collector
+    meta:
+      componentSection: processors
+      signalTypes: [logs]
+      collectorComponentName: transform
+    data:
+      - key: "{{ .ComponentName }}.error_mode"
+        value: "ignore"
+      - key: "{{ .ComponentName }}.log_statements[0]"
+        value: "merge_maps(attributes, attributes[\"metadata\"], \"upsert\") where IsMap(attributes[\"metadata\"])"
```
TylerHelmuth, line 36: This is another situation where the component is making an assumption that `attributes[\"metadata\"]` is gonna exist and have the interesting information we want. I think that is too big of an assumption to make.

TylerHelmuth: I think we should talk with Jessica again about if this specific of a component  is desired. The transform config these concepts came from is really specific to processing the k8sobjectreceiver's kubernetes logs. I dont think it would translate well to a generic solution for a customer.

MikeGoldsmith: I've removed the map merge thing, which can be handled in another processor (see #216). Then updated this processor to specifically only apply otel semconv on k8s object receiver logs.

The other processors I've added recently are more generic, while this one is specific. Do you think that works better?

TylerHelmuth: I feel like this is hyper-specific to the k8sobjectreceiver's events. I've not seen these attributes be extracted in other scenarios as they are commonly added via the k8sattributesprocessor.

PR #220, opened 2025-09-05
feat: Add missing advanced properties to EMA dynamic sampler

MikeGoldsmith:

Updates the EMA dynamic sampler to add missing properties that are useful.

Builder view:

image

Generated config:

```yaml
RulesVersion: 2
Samplers:
    __default__:
        EMADynamicSampler:
            GoalSampleRate: 100
            AdjustmentInterval: 1m0s
            Weight: 0.5
            AgeOutValue: 0.5
            BurstMultiple: 2
            BurstDetectionDelay: 3
            FieldList:
                - http.status_code
            MaxKeys: 500
```

- Add new properties to the sampler
- Update tests to verify expected behaviour

### DIFF
--- a/pkg/data/components/EMADynamicSampler.yaml
+++ b/pkg/data/components/EMADynamicSampler.yaml
@@ -67,6 +67,69 @@ properties:
       - noblanks
       - nonempty
     default: []
+  - name: MaxKeys
+    summary: The maximum number of distinct keys the sampler tracks
+    description: |
+      The maximum number of distinct keys the sampler tracks. This is used to limit
+      the number of keys that are retained in the sampler's state that is used to
+      compute the sampling rate.
+      The default value is 500.
+    type: int
+    validations:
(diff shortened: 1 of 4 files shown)


kentquirk:
I would love for us to use `suppress_if` for the less common properties so that we don't always write all the values to the Refinery config. That complicates matters when Refinery's defaults need to change.

MikeGoldsmith:
> I would love for us to use `suppress_if` for the less common properties so that we don't always write all the values to the Refinery config. That complicates matters when Refinery's defaults need to change.

Yeah, I did consider it and had a version that applied those but had the same thought about if the default changed. I can add the suppress_if's. I'll do the same for the other sampler that was already merged too.

kentquirk (approved):
I like this better. Thanks for making that change.

codeboten (approved):
🎉

PR #226, opened 2025-09-10
feat: Add component validation system with 5 validation types

MikeGoldsmith:
## Summary

This PR implements a comprehensive component validation system for HPSF that validates relationships between multiple component properties. The system supports 5 validation types:

- **at_least_one_of**: Ensures at least one property from a group is provided
- **exactly_one_of**: Ensures exactly one property from a group is provided  
- **mutually_exclusive**: Ensures at most one property from a group is provided
- **require_together**: Ensures all properties in a group are either all empty or all non-empty
- **conditional_require_together**: Ensures properties are non-empty when a condition is met

## Key Changes

- Added `ComponentValidation` struct and `Validations` field to `TemplateComponent`
- Implemented validation logic in `pkg/config/componentValidation.go`
- Integrated validation into translator pipeline after individual property validation
- Added comprehensive unit tests covering all validation types and edge cases
- Added integration tests verifying end-to-end functionality
- Enhanced Transform Processor with `at_least_one_of` validation for statement properties
- Added design documentation and specification files

## Test Plan

- [x] Unit tests for all 5 validation types with various scenarios
- [x] Integration tests through translator validation pipeline  
- [x] Transform Processor validation functionality
- [x] Backward compatibility with components without validations
- [x] Error handling for malformed validations
- [x] Property emptiness semantics (strings, bools, numbers, arrays, maps)

🤖 Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/pkg/config/templateComponent.go
+++ b/pkg/config/templateComponent.go
@@ -53,6 +53,15 @@ type TemplateData struct {
 	Data   []any
 }
 
+// ComponentValidation represents a validation rule that applies to multiple properties
+type ComponentValidation struct {
+	Type              string   `yaml:"type"`
+	Properties        []string `yaml:"properties"`
+	Message           string   `yaml:"message"`
+	ConditionProperty string   `yaml:"condition_property,omitempty"`
+	ConditionValue    any      `yaml:"condition_value,omitempty"`
+}
+
 type ComponentType string
 
 const (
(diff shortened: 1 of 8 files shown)


pkg/config/templateComponent.go
```diff
@@ -53,6 +53,15 @@ type TemplateData struct {
 	Data   []any
 }
 
+// ComponentValidation represents a validation rule that applies to multiple properties
+type ComponentValidation struct {
```
kentquirk, line 57: This is not consistent with the way it's handled at the property level. Those take an optional list of comma-separated arguments, and we simply construct the error message by naming the validation that failed. I think we should do the same here, for usage consistency when writing components, as well as client-side code consistency (we depend on that for constructing the user-visible error message).

kentquirk (commented):
This is definitely in the right spirit. But I do think the style should be consistent with the property validations, for the reasons stated.

PR #229, opened 2025-09-20
fix: Update StartsWith component operator

MikeGoldsmith:

Fixes a bug with the StartsWith Refinery condition component which used an invalid operator value.

- Update StartsWith component value from `startsWith` to `starts-with`
- Use the StartsWith condition to the everything alpha smoke test to validate refinery can start successfully


kentquirk:
Thanks, Mike!

PR #263, opened 2025-12-09
maint: Move Check for Errors condition to deprecated

MikeGoldsmith:

Updates the "Check for Errors" sampling condition to deprecated.

Also updates a test which checks component status when calling TestLoadEmbeddedComponents. We want to be able to load all components regardless of status. The test now verifies that every component has a valid status.

- update component status from alpha to deprecated
- updates TestLoadEmbeddedComponents test to allow deprecated and archived components


TylerHelmuth (approved):

verythorough (changes requested):
@MikeGoldsmith Deprecating this requires having a replacement ready. We technically do with FieldExistsCondition, but as noted in the [issue](https://linear.app/honeycombio/issue/PIPE-269/update-alpha-component-check-if-field-exists#comment-854a40a4) for updating that component, we may have a breaking change there, as well. If so, we should hold off deprecating this one until we have the new version of the other one completed.

(I marked this as changes requested so it didn't get merged accidentally.)

TylerHelmuth (approved):

PR #305, opened 2026-01-20
fix: enforce collector pipelines have a receiver and exporter

TylerHelmuth:

- Fixes an issue where HPSF could render collector config that contained a pipeline without a receiver or exporter. Such a config would not start

Closes PIPE-343

- Add error when rendering the collector config if a pipeline has no receivers or exporters
- Add new function to Validate to enforce OTEL paths starting with a receiver and ending with a exporter or startsampler.
- Add test cases

### DIFF
--- a/pkg/config/tmpl/collectorconfig.go
+++ b/pkg/config/tmpl/collectorconfig.go
@@ -1,6 +1,7 @@
 package tmpl
 
 import (
+	"errors"
 	"fmt"
 	"strings"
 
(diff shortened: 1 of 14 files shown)


pkg/config/tmpl/collectorconfig.go
```diff
@@ -105,6 +105,24 @@ func dedup[T comparable](slice []T) []T {
...
 
+// validatePipelines checks that all collector pipelines have at least one receiver and one exporter
+func (f *collectorConfigFormat) validatePipelines() error {
+	if f.Service == nil {
+		return nil
+	}
+
+	for pipelineName, pipeline := range f.Service.Pipelines {
+		if len(pipeline.Receivers) == 0 {
+			return fmt.Errorf("pipeline '%s' must have at least one receiver", pipelineName)
```
MikeGoldsmith: What do you think about creating an error for each pipeline and combining?

MikeGoldsmith (approved):
Looks good.

Left one non-blocking suggestion about creating an error per pipeline and combing.

